PR #41, opened 2020-01-07
Add young gen RCA and unit tests

rguo-aws:
*Issue #, if available:*

*Description of changes:*
Add young gen RCA and unit tests

*Tests:*
tested on docker image and unit test

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/HighHeapUsageClusterRca.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/HighHeapUsageClusterRca.java
@@ -49,7 +49,7 @@ import org.apache.logging.log4j.Logger;
  */
 public class HighHeapUsageClusterRca extends Rca<HighHeapUsageClusterFlowUnit> {
 
-  private static final Logger LOG = LogManager.getLogger(HighHeapUsageRca.class);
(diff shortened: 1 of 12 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/HighHeapUsageClusterRca.java
```diff
@@ -49,7 +49,7 @@
  */
 public class HighHeapUsageClusterRca extends Rca<HighHeapUsageClusterFlowUnit> {
 
-  private static final Logger LOG = LogManager.getLogger(HighHeapUsageRca.class);
+  private static final Logger LOG = LogManager.getLogger(HighHeapUsageOldGenRca.class);
```
ktkrg: Did you mean `HighHeapUsageClusterRca.class` instead of `HighHeapUsageOldGenRca.class`?

rguo-aws: No, I am trying to replace HighHeapUsageRca with HighHeapUsageOldGenRca because we have both young gen and old gen RCAs.Since we are adding HighHeapUsageYoungGenRca in this PR so it would be better to change the HighHeapUsageRca as well.

ktkrg: Discussed offline.

PR #47, opened 2020-01-10
Make _cat/master call only when rca is enabled

spardeepsingh:
Make _cat/master call only when rca is enabled

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
@@ -260,6 +260,7 @@ public class RcaController {
 
   private String getElectedMasterHostAddress() {
     try {
+      LOG.info("Making _cat/master call");
       final URL url = new URL(CAT_MASTER_URL);
       HttpURLConnection connection = (HttpURLConnection) url.openConnection();
       connection.setRequestMethod("GET");
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
@@ -154,6 +154,7 @@ public class RcaController {
     pollingExecutors = new ArrayList<>();
     pollingExecutors.add(startRcaConfPoller());
     pollingExecutors.add(startRcaNanny());
+    pollingExecutors.add(startNodeRolePoller());
     startExceptionHandlers(pollingExecutors);
   }
 
@@ -214,12 +215,16 @@ public class RcaController {
         this::readRcaEnabledFromConf, 0, pollerPeriodicity, timeUnit);
   }
 
-  private void nodeRolePoller() {
-    final String electedMasterAddress = getElectedMasterHostAddress();
-    final NodeDetails nodeDetails = ClusterDetailsEventProcessor.getCurrentNodeDetails();
-    if (nodeDetails != null) {
-      handleNodeRoleChange(nodeDetails, electedMasterAddress);
-    }
+  private ScheduledFuture<?> startNodeRolePoller() {
+    return netOpsExecutorService.scheduleAtFixedRate(() -> {
+      if (rcaEnabled) {
+        final String electedMasterAddress = getElectedMasterHostAddress();
+        final NodeDetails nodeDetails = ClusterDetailsEventProcessor.getCurrentNodeDetails();
+        if (nodeDetails != null) {
+          handleNodeRoleChange(nodeDetails, electedMasterAddress);
+        }
+      }
+    }, 2, 60, TimeUnit.SECONDS);
   }
 
   /**
@@ -232,7 +237,6 @@ public class RcaController {
     return netOpsExecutorService.scheduleAtFixedRate(
         () -> {
           if (rcaEnabled) {
-            netOpsExecutorService.schedule(this::nodeRolePoller, 2, TimeUnit.SECONDS);
             if (rcaScheduler != null && rcaScheduler.isRunning()) {
               subscriptionManager.dumpStats();
               if (rcaScheduler.getRole() != currentRole) {
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
@@ -218,10 +218,9 @@ public class RcaController {
   private ScheduledFuture<?> startNodeRolePoller() {
     return netOpsExecutorService.scheduleAtFixedRate(() -> {
       if (rcaEnabled) {
-        final String electedMasterAddress = getElectedMasterHostAddress();
         final NodeDetails nodeDetails = ClusterDetailsEventProcessor.getCurrentNodeDetails();
         if (nodeDetails != null) {
-          handleNodeRoleChange(nodeDetails, electedMasterAddress);
+          handleNodeRoleChange(nodeDetails);
         }
       }
     }, 2, 60, TimeUnit.SECONDS);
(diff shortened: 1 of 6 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -249,22 +228,26 @@ private void readRcaEnabledFromConf() {
    * has changed in the meantime (such as a new elected master). It also starts the RCA runtime if
    * it wasn't already running but the current state of the flag expects it to.
    */
-  private void startRcaNanny() {
+  public void startNanny() {
+    startNodeRolePoller();
```
yojs: I tend to think, it we should move it inside the `startPollers()`, so now just two instead of three

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
```diff
@@ -122,17 +122,19 @@ public static NodeDetails getCurrentNodeDetails() {
   }
 
   public static class NodeDetails {
-
     private String id;
     private String hostAddress;
     private String role;
+    private Boolean isMasterNode;
```
yojs: can we use boolean ?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaControllerTest.java
```diff
@@ -30,6 +30,8 @@
 
 @Category(GradleTaskForRca.class)
 public class RcaControllerTest {
+  private static final int IEVAL_ITERATIONS = 20;
+  private static final int NODE_ROLE_EVAL_ITERATIONS = 70;
```
yojs: Is it still required after merging with the last changes ? It never failed for me. If we think it is required, then we should add a comment explaining why this is so

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -133,19 +133,6 @@ public RcaController(
...
 
-  /**
-   * Starts the pollers. Each poller is a thread that checks for the state of the system that the
-   * RCA is concerned with. - RcaConfPoller: Periodically reads a config file to determine if RCA is
-   * supposed to be running or not and accordingly sets a flag. - NodeRolePoller: This checks for
-   * the change of role for an elastic search Master node. - RcaNanny: RcaConfPoller sets the flag
-   * but this thread works to start Rca or shut it down based on the flag.
-   */
-  public void startPollers() {
-    startRcaConfPoller();
```
yojs: It might be cleaner to keep this method and only remove the `startRcaConfPoller();` if we are incorporating that functionality in the `rcaNanny()`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -249,22 +225,26 @@ private void readRcaEnabledFromConf() {
...
-  private void startRcaNanny() {
+  public void startNanny() {
     netOpsExecutorService.scheduleAtFixedRate(
         () -> {
-          if (rcaScheduler != null && rcaScheduler.isRunning()) {
-            if (!rcaEnabled) {
-              // Need to shutdown the rca scheduler
-              stop();
-            } else {
+          netOpsExecutorService.schedule(this::readRcaEnabledFromConf, 1, TimeUnit.SECONDS);
```
yojs: Let's not have calls to `schedule` inside `scheduleAtFixedRate`. We should have just the method references like it used to be. The expensive part was not that the another executor existed (they all use the same thread pool), but the call to `getElectedMasterHostAddress`.  So, let's just make the `nodeRolePoller` a no-op, is RCA is disabled.

yojs:
This is a critical performance bug fix. My two cents would be that we only include the minimum lines of change that are absolutely related to the fixing the bug, which I gain from looking at the PR are two:
- increase the node poller frequency to 60 seconds instead of the old value of 5 seconds.
- Ask for the node role only if RCA is enabled.

I would like to request that only these (and something critically important that I missed) as part of this PR. This will help in a very focused region of interest while doing the review.

It might be good to table the other changes for a subsequent code organization PR as we would like to have more discussion on that. And as they are not pressing need of the hour. 
@spardeepsingh  what are your thoughts ?

ditac:
> This is a critical performance bug fix. My two cents would be that we only include the minimum lines of change that are absolutely related to the fixing the bug, which I gain from looking at the PR are two:
> 
> * increase the node poller frequency to 60 seconds instead of the old value of 5 seconds.
> * Ask for the node role only if RCA is enabled.
> 
> I would like to request that only these (and something critically important that I missed) as part of this PR. This will help in a very focused region of interest while doing the review.
> 
> It might be good to table the other changes for a subsequent code organization PR as we would like to have more discussion on that. And as they are not pressing need of the hour.
> @spardeepsingh what are your thoughts ?

The other part is the change in 

> This is a critical performance bug fix. My two cents would be that we only include the minimum lines of change that are absolutely related to the fixing the bug, which I gain from looking at the PR are two:
> 
> * increase the node poller frequency to 60 seconds instead of the old value of 5 seconds.
> * Ask for the node role only if RCA is enabled.
> 
> I would like to request that only these (and something critically important that I missed) as part of this PR. This will help in a very focused region of interest while doing the review.
> 
> It might be good to table the other changes for a subsequent code organization PR as we would like to have more discussion on that. And as they are not pressing need of the hour.
> @spardeepsingh what are your thoughts ?

There is also the change in ClusterDetailsEventProcessor to read from the EventLog file when master state is available. Can we keep that change as well? It seems pretty separate and self contained to me.

yojs:
@spardeepsingh  I created an issues #50  for this. Please add to it the backwards incompatibility problem we will run into and how your change takes care of it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
spardeepsingh: This value can even be null so we will need Boolean

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaControllerTest.java
spardeepsingh: It is required as we are changing the periodicity of getting master node value to 60 sec. Will add a comment for that

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -277,13 +276,14 @@ private String getElectedMasterHostAddress() {
...
     final NodeRole currentNodeRole = NodeRole.valueOf(currentNode.getRole());
-    if (currentNode.getHostAddress().equalsIgnoreCase(electedMasterHostAddress)) {
-      currentRole = NodeRole.ELECTED_MASTER;
+    Boolean isMasterNode = currentNode.getIsMasterNode();
+    if (isMasterNode != null) {
+      currentRole = isMasterNode ? NodeRole.ELECTED_MASTER : currentNodeRole;
     } else {
-      currentRole = currentNodeRole;
+      final String electedMasterHostAddress = getElectedMasterHostAddress();
+      currentRole = currentNode.getHostAddress().equalsIgnoreCase(electedMasterHostAddress) ? NodeRole.ELECTED_MASTER : currentNodeRole;
```
yojs: Because this is going to make ES API call, perhaps we should have a log for now and then we can add a metric for this.

PR #53, opened 2020-01-13
Use reflection to pass resource type Enum on gPRC

rguo-aws:
*Issue #, if available:*

*Description of changes:*
1. Created a new ResourceType protobuf message and use it to carry the resource type information between nodes. previously we only pass the string name of the enum which is not sufficient especially if we want to do aggregation on master node(i.e. check whether the usage is balanced on some certain resource type)

2. Completely remove the usaeg of apache common "pair" in RCA and make the SlideWindow more generic. Remove the sliding window sub class that was defined in HighHeapOldGenRca.java

3. remove evaluationIntervalSeconds in Rca constructor and hardcoded it to 5 seconds in all Rca. Will raise a subsequent PR to completely get rid of evaluationIntervalSeconds in Rca framework

4. add a boolean variable alwaysCreateSummary in young gen and old gen Rca class. if alwaysCreateSummary is true, the RCA will always create a summary for this flowunit regardless of its state(whether healthy or unhealthy

5. fix some minor issues in unit test. 

*Tests:*
tested on docker image. 


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/HotResourceSummary.java
```diff
@@ -88,8 +109,31 @@ public void buildSummaryMessageAndAddToFlowUnit(FlowUnitMessage.Builder messageB
...
   public static HotResourceSummary buildHotResourceSummaryFromMessage(
       HotResourceSummaryMessage message) {
-    HotResourceSummary newSummary = new HotResourceSummary(message.getResourceType(),
-        message.getThreshold(), message.getValue(), message.getUnitType(), message.getTimePeriod());
+    String resourceTypeClassName = message.getResourceType();
+    //cast to Enum using reflection. find the last "." and separate the class name into two parts.
+    //e.g. com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.Resources$ResourceType.HEAP
+    //will be split to  "com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.Resources$ResourceType"
+    //and "HEAP"
+    String[] nameSplit = resourceTypeClassName.split("\\.(?=[^\\.]+$)");
```
ditac: Can we do this without using reflection? I think its ok to have some boiler plate code here. Ideally, we should define objects that are sent over the network as protobuf objects to avoid this scenario.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/Rca.java
```diff
@@ -24,11 +24,28 @@
 
 public abstract class Rca<T extends ResourceFlowUnit> extends NonLeafNode<T> {
   private static final Logger LOG = LogManager.getLogger(Rca.class);
+  // the amount of RCA period this RCA needs to run before sending out a flowunit
+  protected final int rcaPeriod;
+  protected boolean alwaysCreateSummary;
+  protected int counter;
 
-  public Rca(long evaluationIntervalSeconds) {
+  public Rca(long evaluationIntervalSeconds, int rcaPeriod) {
```
yojs: Going back we have a plan to remove the `evaluationIntervalSeconds` because we think that RCA framework should be simple. It should just run every graph node at every tick. The node decides if it wants 
- to not do any work and return a empty flow unit.
- to do some work and yet return a empty flow unit as it is not time to emit one
- to do some work and emit a flow unit.

So basically the framework runs each graph node each time and it is upto the node to do whatever it wants and return a flow unit empty or otherwise. 
So, we should try to not add the new rcaPeriod

yojs: I would go as far to recommend that we make sure that no RCA depends on `evaluationIntervalSeconds`(by that I mean no RCA requires to be scheduled at a particular interval or depend on that.) as part of this CR and make it vestigial and then in a subsequent CR we can remove it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/HotResourceSummary.java
rguo-aws: Done. Created a ResourceType protobuf message instead and removed the use of  refection/regex from code.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/SlidingWindow.java
```diff
@@ -23,59 +23,74 @@
...
-  protected final int SLIDING_WINDOW_SIZE_IN_SECONDS;
+  protected final Deque<E> windowDeque;
+  protected final long SLIDING_WINDOW_SIZE;
   protected double sum;
 
-  public SlidingWindow(int SLIDING_WINDOW_SIZE_IN_SECONDS) {
+  public SlidingWindow(int SLIDING_WINDOW_SIZE_IN_TIMESTAMP, TimeUnit timeUnit) {
     this.windowDeque = new LinkedList<>();
-    this.SLIDING_WINDOW_SIZE_IN_SECONDS = SLIDING_WINDOW_SIZE_IN_SECONDS;
+    this.SLIDING_WINDOW_SIZE = timeUnit.toSeconds(SLIDING_WINDOW_SIZE_IN_TIMESTAMP);
```
yojs: If we are converting the name unit to seconds, shouldn't the member variable name remain `**_SECONDS` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/Rca.java
rguo-aws: Done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
```diff
@@ -64,7 +64,7 @@ synchronized void createTable(String tableName, List<Field<?>> columns) {
         .column(DSL.field(getPrimaryKeyColumnName(tableName) + PRIMARY_KEY_AUTOINCREMENT_POSTFIX))
         .columns(columns);
 
-    LOG.info("ruizhen: table created: {}", constraintStep.toString());
+    LOG.debug("table created: {}", constraintStep.toString());
```
yojs: We start every log message inside the RCA package as `RCA:` to distinguish the log from the Reader. can we do the same here ?

PR #56, opened 2020-01-14
Fix time taken by RcaController tests

spardeepsingh:
Change Node Role Poller periodicity in tests to be 1 ms

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
@@ -160,9 +160,10 @@ public class PerformanceAnalyzerApp {
             RcaConsts.RCA_CONF_MASTER_PATH,
             RcaConsts.RCA_CONF_IDLE_MASTER_PATH,
             RcaConsts.RCA_CONF_PATH,
-            RcaConsts.networkPollerPeriodicity,
+            RcaConsts.rcaNannyPollerPeriodicity,
+            RcaConsts.rcaConfPollerPeriodicity,
             RcaConsts.nodeRolePollerPeriodicity,
-            RcaConsts.networkPollerPeriodicityTimeUnit);
+            RcaConsts.rcaPollerPeriodicityTimeUnit);
 
     rcaController.startPollers();
   }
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
```diff
@@ -161,6 +161,7 @@ private static void startRcaController(ClientServers clientServers) {
             RcaConsts.RCA_CONF_IDLE_MASTER_PATH,
             RcaConsts.RCA_CONF_PATH,
             RcaConsts.networkPollerPeriodicity,
```
yojs: We can use this opportunity to have a const in RcaCOnst for each poller and name the `networkPollerPeriodicityTimeUnit` as `rcaPollerPeriodicityTimeUnit` ? what do you think ?

yojs:
How long does it take for the build now ?

spardeepsingh:
> How long does it take for the build now ?

It takes less than a minute for the build

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
spardeepsingh: Yeah , we can do that. I just kept same constant for other pollers as they all had same periodicity and if we wanted to change periodicity in future then it will a couple more lines of code change.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -256,8 +256,8 @@ private void removeRcaRequestHandler() {
             }
           }
         },
-        2 * pollerPeriodicity,
-        pollerPeriodicity,
+        2 * rcaNannyPollerPeriodicity,
```
yojs: shouldn't it be `rcaConfPollerPeriodicity` ? As the nanny waits for the first read of the conf ?

spardeepsingh: Yeah you are right, I will fix that

PR #68, opened 2020-01-23
Use WildCardFilter instead of Filename filter to avoid listing all th…

palashhedau:
*Description of changes:* Use WildCardFilter instead of Filename filter to avoid listing all the files in the log directory of rca.sqlite

*Tests:* 
Performance Comparison for different strategy of getting old rca.sqlite file in the log directory
Current Code Performance - Time Elapsed: 11 ms
With NIO DirectoryStream - Time Elapsed: 7 ms
With WildCardFileFilter Elapsed: 5 ms

This has been tested for about 200 files in the folder. 


ditac:
Do we have existing unit test for this code path? Lets add some unit tests to test this particular function call.

PR #86, opened 2020-02-04
handle SQL exception when persisting flowunit

rguo-aws:
*Issue #, if available:*

*Description of changes:*
The current SQL table creation and insertion methods are not handling SQLException properly. This PR is raised to handle SQL exception if found. If the rca.sqlite file is damaged or accidentally deleted, RCA will try to create a new db file immediately and reconnect to this new DB.  

*Tests:*
tested on AES cluster. 

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PersistorBase.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PersistorBase.java
@@ -80,9 +80,9 @@ public abstract class PersistorBase implements Persistable {
       String tableName,
       List<Field<?>> columns,
       String refTable,
-      String referenceTablePrimaryKeyFieldName);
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PersistorBase.java
```diff
@@ -191,31 +191,48 @@ public synchronized void deleteOldDBFile()
...
+        writeSummary(
+            flowUnit.getResourceSummary(),
+            tableName,
+            getPrimaryKeyColumnName(tableName),
+            lastPrimaryKey);
+      }
+    } catch (SQLException e) {
+      LOG.info(
+          "RCA: Fail to write into table '{}', try recreating the DB", tableName);
+      openNewDBFile();
```
yojs: we are opening a new file but we are not trying to write in the new file one more time. We are throwing away this flow unit here

PR #88, opened 2020-02-05
Rca persistence layer refactoring.

yojs:
*Description of changes:*
There are three parts to it: 
- Before this patch, the Rca sqlite files would be created as rca.sqlite<timestamp>. We want to change it to rca.sqlite and add the time stamp only when the file is rotated and no longer a candidate for modification.

- In the code as it stands, we would do a directory scan to identify the old db files that are candidates for deletion. This is performance intensive. This patch adds changes such that a full directory scan is only required at startup and never again.

- Separate out the logic for db file rotation and garbage collection into separate files and out of `PersistorBase`.

*Tests:*
- Added unit tests for test the rotation logic
-  added unit tests for the file garbage collection logic
- added unit tests to check that sqlite files are correctly written, rotated and garbage collected. And the data inside them is flushed to disk and hence readable (no corruption under normal circumstances). We were missing these tests before.

For the files added it covers more than 80%

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/FileRotate.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/FileRotate.java
@@ -43,7 +43,7 @@ public class FileRotate {
       long rotation_period,
       DateFormat rotated_file_format) {
     FILE_TO_ROTATE = file_to_rotate;
-    FILENAME = file_to_rotate.getFileName().toString();
+    FILENAME = file_to_rotate.toFile().getName();
     ROTATION_TIME_UNIT = rotation_time_unit;
     ROTATION_PERIOD = rotation_period;
     ROTATED_FILE_FORMAT = rotated_file_format;
@@ -93,13 +93,17 @@ public class FileRotate {
   /**
    * Rotate the file.
    *
-   * <p>Rotating a file renames it to filename.<current time in the date format specified>. For
+   * <p>Rotating a file renames it to filename.#current time in the date format specified#. For
    * cases the file could not be renamed, we attempt to delete the file. If the file could not be
    * deleted either, we throw an IOException for the caller to handle or take the necessary action.
    *
    * @return Returns the path to the file after it was rotated.
    */
   protected Path rotate(long currentMillis) throws IOException {
+    if (!FILE_TO_ROTATE.toFile().exists() || FILE_TO_ROTATE.getParent() == null) {
+      return null;
+    }
+
     String dir = FILE_TO_ROTATE.getParent().toString();
     StringBuilder targetFileName = new StringBuilder(FILENAME);
     targetFileName.append(FILE_PART_SEPARATOR).append(ROTATED_FILE_FORMAT.format(currentMillis));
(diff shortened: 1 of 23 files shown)


src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaTestHelper.java
```diff
@@ -56,6 +56,31 @@
     return matches;
   }
 
+  public static List<String> getAllLinesFromLog(String logName) {
+    try {
+      return Files.readAllLines(Paths.get(getLogFilePath(logName)));
+    } catch (IOException e) {
```
palashhedau: nit: we can merge all exception using pipe, as we are only printing stacktrace

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/FileRotate.java
```diff
@@ -0,0 +1,129 @@
...
+
+  /**
+   * This checks for the condition for file rotation.
+   *
+   * <p>This usually checks if the time since last rotation is past the {@code ROTATION_PERIOD}.
+   *
+   * @return
+   */
+  protected boolean shouldRotate(long currentTimeMillis) {
+    long timeSinceLastRotation = currentTimeMillis - lastRotatedMillis;
```
palashhedau: Just curious, we are rotating file on the basis of time passed or should we rotate if the current time hour > file hour

for example: at 4:00PM, we should rotate file which is even created at 3.30
It should happen as we are rotating file at 4.30.

Had a discussion regarding this with Karthik

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaTestHelper.java
```diff
@@ -110,4 +114,14 @@ public static void cleanUpLogs() {
       e.printStackTrace();
     }
   }
+
+  public static void truncate(File file) {
+    try (FileChannel outChan = new FileOutputStream(file, false).getChannel()) {
+      outChan.truncate(0);
+    } catch (FileNotFoundException e) {
+      System.out.println(file.getName() + " does not exist.");
```
palashhedau: Sysout or log?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaTestHelper.java
yojs: done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/FileRotate.java
yojs: We are rotating based on 1 hour since the file creation.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaTestHelper.java
yojs: For tests System.out is okay. We don't usually look at logs.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/FileRotate.java
yojs: I will probably add a subsequent patch to have this change. Don't want to make this PR bigger.

PR #90, opened 2020-02-06
RCA persistence framework refactorings.

yojs:
Description of changes:
There are three parts to it:

- Before this patch, the Rca sqlite files would be created as rca.sqlite. We want to change it to rca.sqlite and add the time stamp only when the file is rotated and no longer a candidate for modification.
- In the code as it stands, we would do a directory scan to identify the old db files that are candidates for deletion. This is performance intensive. This patch adds changes such that a full directory scan is only required at startup and never again.
- Separate out the logic for db file rotation and garbage collection into separate files and out of PersistorBase.
- Add the logic to attempt to write data in case of SQL exception after recreating file.

Tests:

- Added unit tests for test the rotation logic
- added unit tests for the file garbage collection logic
- added unit tests to check that sqlite files are correctly written, rotated and garbage collected. And the data inside them is flushed to disk and hence readable (no corruption under normal circumstances). We were missing these tests before.

Code coverage percentage for this patch:
For the files added it covers more than 80%

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/RcaConsts.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/RcaConsts.java
@@ -60,7 +60,7 @@ public class RcaConsts {
   public static final String DATASTORE_STATE_COL_NAME = "state";
   public static final String DATASTORE_STORAGE_FILE_RETENTION_COUNT = "storage-file-retention-count";
 
-  // The next two lines says that that the RCA sqlite files needs to be rotated every hour
+  // The next two lines says that the RCA sqlite files needs to be rotated every hour
   public static final TimeUnit DB_FILE_ROTATION_TIME_UNIT = TimeUnit.HOURS;
   public static final long ROTATION_PERIOD = 1;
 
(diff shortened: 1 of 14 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/RcaConsts.java
```diff
@@ -60,6 +60,10 @@
   public static final String DATASTORE_STATE_COL_NAME = "state";
   public static final String DATASTORE_STORAGE_FILE_RETENTION_COUNT = "storage-file-retention-count";
 
+  // The next two lines says that that the RCA sqlite files needs to be rotated every hour
```
ktkrg: that that -> that

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/FileGC.java
```diff
@@ -0,0 +1,147 @@
...
+        deletedSoFar += 1;
+      } else {
+        remainingFiles.add(file);
+      }
+    }
+    return remainingFiles;
+  }
+
+  private void delete(File file) {
+    if (!file.delete()) {
```
rguo-aws: Do we need to catch exception here ?

yojs: Good catch! we should. Added a revision to throw exception if we fail to delete not just silently logging it

Issue #100, opened 2020-02-12
Fix Rca Sqlite Schema


spardeepsingh:
There is a bug in Rca Sqlite Schema. The table dependency should be in following order:
Rca_Table -> Hot_Cluster_Summary -> Hot_Node_Summary -> Hot_Resource_Summary. 

Currently HotNodeRca table is not linked to Hot_Cluster_Summary. It is linked directly to Hot_Node_Summary which will cause issues when we insert into these tables. We need to link HotNodeRca directly to HotClusterSummary.

We should also rename the foreign_Key column in HotClusterSummary table as Rca_Id and have unique primary keys for all available Rcas

PR #102, opened 2020-02-12
Persist nested summaries only if the flowunit is emitted by the top l…

rguo-aws:
…evel RCA of each node

*Issue #, if available:*
#100 

*Description of changes:*
Add a "persistSummary" flag in flowunit so that only the flowunit emitted by the top level RCA can persist its summary. For data node, the top level RCA is HotNodeRca and for elected master, it is those cluster level RCAs. 
This can help to fix two different issues : 
1. summary will no longer be persisted multiple times. Previously each RCA will persist its own summary and the inner most summary will be persisted by multiple flowunits. 
2. The HotNodeSummary table on master can now can only be linked to HotClusterSummary. This is to avoid the HotNodeSummary table being created and linked to the HotNodeRca table first and causes the RCA API to fail. 

*Tests:*
tested on docker

n/a

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/HotNodeRca.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/HotNodeRca.java
@@ -94,7 +94,7 @@ public class HotNodeRca extends Rca<ResourceFlowUnit> {
       // reset the variables
       counter = 0;
       hasUnhealthyFlowUnit = false;
-      return new ResourceFlowUnit(System.currentTimeMillis(), context, summary);
+      return new ResourceFlowUnit(System.currentTimeMillis(), context, summary, !currentNode.getIsMasterNode());
     } else {
       return new ResourceFlowUnit(System.currentTimeMillis());
     }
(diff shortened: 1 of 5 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/HotNodeRca.java
```diff
@@ -94,7 +94,7 @@ public ResourceFlowUnit operate() {
       // reset the variables
       counter = 0;
       hasUnhealthyFlowUnit = false;
-      return new ResourceFlowUnit(System.currentTimeMillis(), context, summary);
+      return new ResourceFlowUnit(System.currentTimeMillis(), context, summary, !currentNode.getIsMasterNode());
```
yojs: for readability, we might want to add this to a variable with appropriate name and add a line of comment why the masterNode does not need to persist the summary.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistorTest.java
```diff
@@ -94,7 +94,7 @@ public void write() throws IOException, SQLException, InterruptedException {
             71,
             "heap usage in percentage",
             60);
-    ResourceFlowUnit rfu = new ResourceFlowUnit(System.currentTimeMillis(), context, summary);
+    ResourceFlowUnit rfu = new ResourceFlowUnit(System.currentTimeMillis(), context, summary, true);
 
     Node rca = new TestRca();
```
yojs: It might makes sense to add a test case that creates a graph which executes on data node and the master partly and then check if it is persisted at each places. 
We hit the bug because we didn't have a Unit test to validate this in the first place. Let's add one now

spardeepsingh:
Looks good. As we are fixing the schema, can we also change the foreign key column name in HotClusterSummary table to something like Rca_Id, so that it does not randomly pick any name from available rcas?

PR #103, opened 2020-02-13
Persist nested summaries only if the flowunit is emitted by the top level RCA of each node

rguo-aws:
*Issue #, if available:*

*Description of changes:*
Add a "persistSummary" flag in flowunit so that only the flowunit emitted by the top level RCA can persist its summary. For data node, the top level RCA is HotNodeRca and for elected master, it is those cluster level RCAs.
This can help to fix two different issues :

summary will no longer be persisted multiple times. 
1. Previously each RCA will persist its own summary and the inner most summary will be persisted by multiple flowunits.
2. The HotNodeSummary table on master can now can only be linked to HotClusterSummary. This is to avoid the HotNodeSummary table being created and linked to the HotNodeRca table first and causes the RCA API to fail.

*Tests:*
Added a unit test. 
tested on docker

n/a


src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PersistFlowUnitAndSummaryTest.java
```diff
@@ -0,0 +1,206 @@
...
+  Queryable queryable;
+
+  static class DummyYoungGenRca extends Rca<ResourceFlowUnit> {
+    public <M extends Metric> DummyYoungGenRca(M metric) {
+      super(1);
+    }
+
+    @Override
+    public ResourceFlowUnit operate() {
+      System.out.println("!!!!! ruizhen : scheduled \n");
```
yojs: probably we want to remove this ? :)

rguo-aws: done

Issue #105, opened 2020-02-13
Refactor threading framework


ktkrg:
### Overview
Currently the PA engine process spins up multiple threads in different places in order to run the reader, and the RCA. We need to consolidate the these executor/thread creation, rethink the execution hierarchy, and handle missed cases of error handling.

When Performance Analyzer engine is started, it should spawn a top level thread(or reuse the main thread as the top level thread) to then spin up a new thread for the following concurrent activities:

1. PA Reader
1. gRPC server
1. HTTP/1.x server for cURL and perf-top usecases
1. RCA controller

### Top level thread functionality
The top level thread spawns the necessary high level secondary threads. Once that is done, it just blocks on the error queue and waits for an exception to be put in the queue. If it sees an error in the queue, it dequeues the error, handles the error appropriately and continues to block on the queue for the lifetime of the agent.

### Thread provisioning
Whenever a new second-level thread is needed, the top level thread also passes in an callback to the second level runnable, so that the child thread can signal the top level thread if it runs into any error that it wishes to notify the parent.

The top level thread also uses the information from provisioning for accounting purposes.

### Consolidation of RCA executors
This only captures the high level idea and not the specifics.

We need to move the scheduler off of the ScheduledExecutorService into its own thread which again, runs in a loop, evaluates the graph and sleeps for the remainder of the second if it finishes early.

The multi-nanny executor can be run as part of the second-level RCA controller itself instead of it having its own executor and separate error handling threads.

**_PHASE 1_**
* Create top level controller
* Provision the above threads from the top level controller

**Testing**
Broadly speaking all of the following components need to be tested
* Reader tests
  * Ensure reader is working as expected, creating db files, emitting stats(metrics) in all configurations
  * [Placeholder: Tests for error conditions the previous implementation would've missed]
* RCA tests
  * Ensure RCA is working as expected, enable/disable, setting up and tearing down of executors, and error handling threads.
  * [Placeholder: Tests for error conditions the previous implementation would've missed]
* perftop and curl tests
  * Ensure curl'd requests return valid responses, perf-top dashboards are showing all metrics.
  * [Placeholder: Tests for error conditions the previous implementation would've missed]

yojs:
`the top level thread also passes in an listener to ..` should this be callback ?

In the  "Consolidation of RCA executors", we should probably mention the rationale so that people understand why ?

[...]

PR #108, opened 2020-02-21
Manage threads explicitly instead of using executors - 1/3

ktkrg:
*Issue #, if available:*
#105 
*Description of changes:*
1. Create a singleton thread provider which provisions threads and handles exceptions by writing to a queue. (The queue is currently a dummy placeholder for correct error handling)
2. Manage the RcaScheduler thread ourselves instead of using an executor and a separate exception handling thread. This exception handling thread led to the spike in master CPU utilization.
3. Manage the RcaController as a separate thread instead of using 3 scheduled tasks. It removes creation of 3 exception handling threads as well.

With all of the changes, there is no change in behavior other than the fact we don't see a spike in master CPU utilization. We can improve the error handling with per-task error handling if need be in the next set of commits.

All threads die if they throw an exception that the thread itself is not catching and this leads to that thread dying.
*Tests:*
Tested on opendistro, unit tests pass.


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
```diff
@@ -42,36 +41,30 @@
...
-import java.net.InetSocketAddress;
-import java.security.KeyStore;
-import java.security.Security;
-import java.security.cert.X509Certificate;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
```
ditac: Lets use ArrayBlockingQueue

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tasks/ThreadProvider.java
```diff
@@ -0,0 +1,47 @@
...
+    _instance = new ThreadProvider();
+    return _instance;
+  }
+
+  public Thread createThreadForRunnable(final Runnable innerRunnable, final String name) {
+    Thread t = new Thread(() -> {
+      try {
+        innerRunnable.run();
+      } catch (Throwable innerThrowable) {
+        if (!PerformanceAnalyzerApp.exceptionQueue.offer(new PAThreadException(name,
```
ditac: We should just block if the queue is full. We should emit metrics on errors and these metrics need to be accurate. The component will be dead if it hits an exception. Should we restart the process if we run into exceptions?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
ktkrg: Changed it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
```diff
@@ -42,36 +41,30 @@
...
 
 public class PerformanceAnalyzerApp {
+
   private static final int WEBSERVICE_DEFAULT_PORT = 9600;
   private static final String WEBSERVICE_PORT_CONF_NAME = "webservice-listener-port";
   private static final String WEBSERVICE_BIND_HOST_NAME = "webservice-bind-host";
   // Use system default for max backlog.
   private static final int INCOMING_QUEUE_LENGTH = 1;
+  // current number of threads spawned through the thread provider.
+  private static final int EXCEPTION_QUEUE_LENGTH = 5;
```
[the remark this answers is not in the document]
ktkrg: I've changed the queue size to one, and with ArrayBlockingQueue, we block till we process the exception.

The previous idea was to poll on the queue with a specific periodicity, and since we were spawning 5 threads(1 reader thread, 1 rca controller, 2 server threads) the queue length of 5 would let all of them to write to the queue without blocking within one polling interval of the exception queue.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
```diff
@@ -84,244 +77,147 @@
   private static Thread rcaNetServerThread = null;
 
   public static final SampleAggregator RCA_GRAPH_METRICS_AGGREGATOR =
-          new SampleAggregator(RcaGraphMetrics.values());
-  public  static final SampleAggregator RCA_RUNTIME_METRICS_AGGREGATOR =
-          new SampleAggregator(RcaRuntimeMetrics.values());
+      new SampleAggregator(RcaGraphMetrics.values());
```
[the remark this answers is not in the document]
ktkrg: Looks like it was previously not (strictly) formatted according to the google style guide. Now it adheres to the style guide correctly which is enforced by the formatter in intellij. The continuation indent parameter which controls this is defined as follows:
> "When line-wrapping, each line after the first (each continuation line) is indented at least +4 from the original line. When there are multiple continuation lines, indentation may be varied beyond +4 as desired."

instead of varying the space beyond +4, it is strictly setting it to exactly +4 now.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
```diff
@@ -84,244 +77,147 @@
...
-        server = createHttpsServer(internalPort);
-      } else {
-        server = createHttpServer(internalPort);
+  private static void startErrorHandlingThread() {
+    final Thread errorHandlingThread = ThreadProvider.instance().createThreadForRunnable(() -> {
+      while (true) {
+        try {
+          long startTime = System.currentTimeMillis();
+          List<PAThreadException> exceptions = new ArrayList<>();
+          exceptionQueue.drainTo(exceptions);
```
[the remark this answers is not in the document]
ktkrg: Doing that now. Waiting on exceptions with a take().

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tasks/ThreadProvider.java
```diff
@@ -0,0 +1,47 @@
...
+  private static volatile ThreadProvider _instance = null;
+  private int numberOfThreadsSpunUp = 0;
+
+  /**
+   * Empty private default ctor to prevent instantiation.
+   */
+  private ThreadProvider() {
+  }
+
+  public static ThreadProvider instance() {
```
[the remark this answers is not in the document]
ktkrg: Got rid of the singleton. It is initialized once and passed around now.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tasks/ThreadProvider.java
ktkrg: Blocking on the queue with put().

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tasks/ThreadProvider.java
```diff
@@ -0,0 +1,47 @@
...
+        if (!PerformanceAnalyzerApp.exceptionQueue.offer(new PAThreadException(name,
+            innerThrowable))) {
+          LOG.error("Unable to write to exception queue. Dropping exception: {}",
+              innerThrowable.getMessage(), innerThrowable.getCause());
+        }
+      }
+    });
+
+    LOG.info("Spun up a thread with name: {}", name);
+    this.numberOfThreadsSpunUp++;
```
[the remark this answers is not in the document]
ktkrg: got rid of the unused variable. Instead, made this method synchronized and emitting a metric from it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/threads/ThreadProvider.java
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * Creates a thread which executes the given runnable when started. If the given runnable throws
+   * an uncaught exception, it is then written to the exception queue which will be processed by the
+   * exception handler thread.
+   *
+   * @param innerRunnable The runnable to execute when the thread starts.
+   * @param paThread      The thread enum value from {@link PerformanceAnalyzerThreads}
+   * @return The thread with the wrapped runnable.
+   */
+  public synchronized Thread createThreadForRunnable(final Runnable innerRunnable,
```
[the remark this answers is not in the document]
ktkrg: Removed it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaControllerHelper.java
```diff
@@ -0,0 +1,108 @@
...
+   */
+  public static ExecutorService buildNetworkThreadPool(final int queueLength) {
+    final ThreadFactory rcaNetThreadFactory =
+        new ThreadFactoryBuilder().setNameFormat(RcaConsts.RCA_NETWORK_THREAD_NAME_FORMAT)
+                                  .setDaemon(true)
+                                  .build();
+    final BlockingQueue<Runnable> threadPoolQueue = new LinkedBlockingQueue<>(queueLength);
+    return new ThreadPoolExecutor(RcaConsts.NETWORK_CORE_THREAD_COUNT,
+        RcaConsts.NETWORK_MAX_THREAD_COUNT, 0L, TimeUnit.MILLISECONDS, threadPoolQueue,
+        rcaNetThreadFactory);
```
yojs: I was thinking that we move the entire thread creation into a helper class. The thread creators should use that to create threads using few parameters

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/RcaConsts.java
```diff
@@ -66,7 +67,7 @@
 
   public static final long rcaNannyPollerPeriodicity = 5;
   public static final long rcaConfPollerPeriodicity = 5;
-  public static final long nodeRolePollerPeriodicity = 60;
+  public static final long nodeRolePollerPeriodicityInSeconds = 60;
```
yojs: screaming snake case for final ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/scheduler/RCAScheduler.java
```diff
@@ -72,6 +65,7 @@
   ThresholdMain thresholdMain;
   Persistable persistable;
   static final int PERIODICITY_SECONDS = 1;
+  static final int PERIODICITY_IN_MS = PERIODICITY_SECONDS * 1000;
   ScheduledFuture<?> futureHandle;
```
yojs: I see that the users of the field are deleted. we should remove this too ? Did I miss something ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/scheduler/RCAScheduler.java
```diff
@@ -99,52 +93,30 @@ public void start() {
...
+          task.run();
+          long duration = System.currentTimeMillis() - startTime;
+          if (duration < PERIODICITY_IN_MS) {
+            Thread.sleep(PERIODICITY_IN_MS - duration);
+          }
+        } catch (InterruptedException ie) {
+          LOG.error("Rca scheduler thread sleep interrupted. Reason: {}", ie.getMessage());
+          LOG.error(ie);
+          shutdown();
+          schedulerState = RcaSchedulerState.STATE_STOPPED_DUE_TO_EXCEPTION;
```
yojs: We are catching exception here. How are we reporting that a thread shutdown because of exception in the stats collector ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/threads/ThreadProvider.java
```diff
@@ -0,0 +1,65 @@
...
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.PerformanceAnalyzerThreads;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.collectors.StatsCollector;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.threads.exceptions.PAThreadException;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * Class that wraps a given runnable in a thread with exception handling capabilities.
+ */
+public class ThreadProvider {
```
yojs: I was also thinking that this will have a method say `runningThreadCount` and `expectedThreadCount` which will be polled by `PeriodicSamplers`. In normal case these should be equal but in case a thread dies we would know. It should not be a point emission (emit once and never again). It should be emitted until the problem is fixed. We should be able to see in the fleet if there are threads that are dead now and requires an Agent restart.

yojs:
@ktkrg I have some comments on this PR. Let's discuss if you agree and maybe we can create another PR to fix ?

PR #125, opened 2020-03-10
Create a base class RCA ResourceResourceRca as a generic resource type RCA. Add HighCpuRca to collect total cpu usage

rguo-aws:
*Issue #, if available:*

*Description of changes:*
The GenericResourceRca is a base type RCA class which collect aggregated metric from a certain resource type and calculated its averaged sum by using a sliding window.

The HighCpuRca derives from the TotalResourceRca to calculate the cpu usage of its node and list the top k consumers.

Add a unit test for this rca class

*Tests:*
tested on dev stack 

82%


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/hot_node/GenericResourceRca.java
```diff
@@ -0,0 +1,201 @@
...
+  private static final Logger LOG = LogManager.getLogger(GenericResourceRca.class);
+  private static final int SLIDING_WINDOW_IN_MIN = 10;
+  private static final int DEFAULT_TOP_K = 3;
+  private static final double DEFAULT_LOWER_BOUND_THRESHOLD = 0.0;
+  private final SlidingWindow<SlidingWindowData> slidingWindow;
+  private final Metric resourceUsageGroupByConsumer;
+  private final int rcaPeriod;
+  private int counter;
+  private final ResourceType resourceType;
+  private double threshold;
```
ditac: Can we make this final so that its threadsafe?

rguo-aws: I have changed those variables to volatile to make them thread safe

PR #133, opened 2020-04-07
Add unit tests to the rca.framework.api.persist package

sidheart:
*Issue #, if available:* 51

*Description of changes:* Add unit tests to the rca.framework.api.persist package 

*Tests:* SQLParsingUtilTest - verifies the behavior of the SQLParsingUtil

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
@@ -37,14 +37,15 @@ public class SQLParsingUtil {
    * @throws IllegalArgumentException throws an exception if we can't find the name in the field
    */
   private static Record getRecordByName(Result<Record> result, Field<String> field, String name) throws IllegalArgumentException {
+    if (result == null) {
+      throw new IllegalArgumentException("getRecordByName called with null result");
+    }
     int idx = -1;
-    if (result != null) {
-      List<String> fieldNames = result.getValues(field);
-      for (int i = 0; i < fieldNames.size(); i++) {
-        if (fieldNames.get(i).equals(name)) {
-          idx = i;
-          break;
-        }
+    List<String> fieldNames = result.getValues(field);
+    for (int i = 0; i < fieldNames.size(); i++) {
+      if (fieldNames.get(i).equals(name)) {
+        idx = i;
+        break;
       }
     }
     if (idx < 0) {
@@ -70,18 +71,15 @@ public class SQLParsingUtil {
    */
   public static double readDataFromSqlResult(Result<Record> result, Field<String> matchedField, String matchedFieldName, String dataField) {
     double ret = Double.NaN;
-    if (result != null) {
-      try {
-        Record record = SQLParsingUtil.getRecordByName(result, matchedField, matchedFieldName);
-        ret = record.getValue(MetricsDB.MAX, Double.class);
-      }
-      catch (IllegalArgumentException ie) {
-        LOG.error("{} fails to match any row in field {}.", matchedFieldName, matchedField.getName());
-      }
-      catch (
-          DataTypeException de) {
-        LOG.error("Fail to read {} field from SQL result", dataField);
-      }
+    try {
+      Record record = getRecordByName(result, matchedField, matchedFieldName);
+      ret = record.getValue(MetricsDB.MAX, Double.class);
+    }
+    catch (IllegalArgumentException ie) {
+      LOG.error("{} fails to match any row in field {}.", matchedFieldName, matchedField.getName());
+    }
+    catch (DataTypeException de) {
+      LOG.error("Fail to read {} field from SQL result", dataField);
     }
     return ret;
   }
--- a/src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtilTest.java
+++ b/src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtilTest.java
@@ -1,0 +1,55 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.persist;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.when;
+
+import com.google.common.collect.Lists;
+import org.jooq.Field;
+import org.jooq.Record;
+import org.jooq.Result;
+import org.jooq.exception.DataTypeException;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+public class SQLParsingUtilTest {
+    private static final String FIELD_NAME = "FIELD_NAME";
+    private static final String DATA_FIELD = "DATA_FIELD";
+
+    @Mock
+    private Result<Record> result;
+
+    @Mock
+    private Field<String> field;
+
+    @Mock
+    private Record record;
+
+    @Before
+    public void setup() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void testReadDataFromSqlResult() {
+        // method should return Double.NaN when result is null
+        assertEquals(Double.NaN, SQLParsingUtil.readDataFromSqlResult(null, field, FIELD_NAME, DATA_FIELD), 0);
+
+        // method should return Double.NaN when it encounters an IllegalArgumentException
+        when(result.getValues(field)).thenReturn(Lists.newArrayList("no", "matches", "here"));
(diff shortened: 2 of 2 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
```diff
@@ -68,11 +75,11 @@ private static Record getRecordByName(Result<Record> result, Field<String> field
    * @param dataField The field to retrieve data from
    * @return the data read from SQL result
    */
-  public static double readDataFromSqlResult(Result<Record> result, Field<String> matchedField, String matchedFieldName, String dataField) {
+  public double readDataFromSqlResult(Result<Record> result, Field<String> matchedField, String matchedFieldName, String dataField) {
     double ret = Double.NaN;
     if (result != null) {
       try {
-        Record record = SQLParsingUtil.getRecordByName(result, matchedField, matchedFieldName);
+        Record record = getRecordByName(result, matchedField, matchedFieldName);
```
yojs, line 82: `getRecordByName()` is marked `VisibleForTesting` but this caller is not. When the application (and not the tests) calls it, it will result in error, right ?

sidheart: The @VisibleForTesting annotation just indicates that an element is only used for testing or has wider scope than is strictly necessary to allow for testing. I changed the method signature to be package private so that it could be tested, which is why I included the annotation. The annotation doesn't actually do anything at runtime.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
```diff
@@ -27,6 +28,10 @@
 public class SQLParsingUtil {
   private static final Logger LOG = LogManager.getLogger(SQLParsingUtil.class);
 
+  public static final SQLParsingUtil API = new SQLParsingUtil();
```
rguo-aws, line 31: do we really need singleton pattern here ? I assume those are just some static utility functions which can be tested directly without singleton.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtilTest.java
```diff
@@ -0,0 +1,76 @@
...
+        when(result.getValues(field)).thenReturn(Lists.newArrayList("no", "matches", "here"));
+        uut.getRecordByName(result, field, FIELD_NAME);
+    }
+
+    @Test
+    public void testReadDataFromSqlResult() {
+        // method should return Double.NaN when result is null
+        assertEquals(Double.NaN, uut.readDataFromSqlResult(null, field, FIELD_NAME, DATA_FIELD), 0);
+        // method should return Double.NaN when it encounters an IllegalArgumentException
+        when(result.getValues(field)).thenReturn(Lists.newArrayList("no", "matches", "here"));
```
rguo-aws, line 65: I would suggest to create a dummy SQL context instead of mocking the Result object. You can use the jooq DSLcontext to create whatever Result<Record> you want and use that as input for this test.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
```diff
@@ -36,15 +41,17 @@
    * @return The record that contain the name string in field
    * @throws IllegalArgumentException throws an exception if we can't find the name in the field
    */
-  private static Record getRecordByName(Result<Record> result, Field<String> field, String name) throws IllegalArgumentException {
+  @VisibleForTesting
+  Record getRecordByName(Result<Record> result, Field<String> field, String name) throws IllegalArgumentException {
```
rguo-aws, line 45: not sure if we really need to change the signature of this function from private to public. I assume we can still call the public API readDataFromSqlResult to test this function ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
sidheart: Removed the annotation, returned to static, non-singleton

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
sidheart: Removed, now tested directly

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtilTest.java
sidheart: I disagree that this would be better than my current method of testing with mocks. I also think using the jooq DSL context would increase the overall line count of this test. Is there a strong reason to not use Mockito here?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
sidheart: Changed it back, you're absolutely correct

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtilTest.java
rguo-aws: if we use mockito it would be hard to tell whether the exception "DataTypeException" can be emitted properly if we try to parse the result with wrong field name / data type. Anyway. I think it is a good to have and we can skip it for now.  :)

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
```diff
@@ -37,14 +37,15 @@
    * @throws IllegalArgumentException throws an exception if we can't find the name in the field
    */
   private static Record getRecordByName(Result<Record> result, Field<String> field, String name) throws IllegalArgumentException {
```
yojs, line 39: If we don't expect `result` to be `null`, we should add the notnull annotation for that argument.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
```diff
@@ -70,18 +71,16 @@ private static Record getRecordByName(Result<Record> result, Field<String> field
...
-      }
-      catch (
-          DataTypeException de) {
-        LOG.error("Fail to read {} field from SQL result", dataField);
-      }
+    try {
+      Record record = getRecordByName(result, matchedField, matchedFieldName);
+      ret = record.getValue(MetricsDB.MAX, Double.class);
+    }
+    catch (IllegalArgumentException ie) {
```
yojs, line 78: Can we add a test to check Illegal Argument Exception and DataTypeException is thrown ?

sidheart: Discussed how this is covered

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
```diff
@@ -70,18 +71,16 @@ private static Record getRecordByName(Result<Record> result, Field<String> field
...
+      LOG.error("{} fails to match any row in field {}.", matchedFieldName, matchedField.getName());
+    }
+    catch (
+            DataTypeException de) {
```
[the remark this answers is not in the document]
sidheart: Done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/persist/SQLParsingUtil.java
sidheart: Disagree, we explicitly included logic dealing with null results which results in a logging statement and a particular double.NaN return statement from the public caller. Adding the annotation to this private method would only shift the (result == null) logic to the public function and slightly increase the overall line count.

yojs: Throwing an exception is a runtime validation but I think we should try and prefer static checks over them as this way the static analysis tools can do the work for you before the code goes to prod. Anyways, the @nonnull is not enforced by the compiler. So, we can ignore it.

PR #135, opened 2020-04-07
Added unit tests for rca.samplers and rca.util

sidheart:
*Issue #, if available:* 51

*Description of changes:* 
- Changed visibility of a few variables and methods to allow for testing
- Added ability to create new NodeDetails elements to the ClusterDetailsEventProcessorTestHelper class

*Tests:*
- RcaEnabledSamplerTest: Basic test which verified the construction and sampling behavior of the uut
- Modified NodeStateManagerTest to not rely on mocks
- Added the RcaStateSamplersTest which does basic constructor coverage. We can't ignore this package, so this test is minor, but necessary for improved coverage
- Added the ClusterUtilsTest which verifies ClusterUtils behavior

*Code coverage percentage for this patch:* 93% for rca.samplers
                                                                      89% for rca.util

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/RcaEnabledSampler.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/RcaEnabledSampler.java
@@ -21,20 +21,21 @@ import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.stats.collect
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.stats.emitters.ISampler;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.reader.ClusterDetailsEventProcessor;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.reader.ClusterDetailsEventProcessor.NodeDetails;
+import com.google.common.annotations.VisibleForTesting;
 
 public class RcaEnabledSampler implements ISampler {
 
   @Override
   public void sample(SampleAggregator sampleCollector) {
-    sampleCollector.updateStat(RcaRuntimeMetrics.RCA_ENABLED, "", isRcaEnabled());
+    sampleCollector.updateStat(RcaRuntimeMetrics.RCA_ENABLED, "", isRcaEnabled() ? 1 : 0);
   }
 
-  private int isRcaEnabled() {
+  @VisibleForTesting
+  boolean isRcaEnabled() {
     NodeDetails currentNode = ClusterDetailsEventProcessor.getCurrentNodeDetails();
     if (currentNode != null && currentNode.getIsMasterNode()) {
-      return RcaController.isRcaEnabled() ? 1 : 0;
+      return RcaController.isRcaEnabled();
     }
-
-    return 0;
+    return false;
   }
 }
--- a/src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessorTestHelper.java
+++ b/src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessorTestHelper.java
@@ -15,6 +15,7 @@
 
 package com.amazon.opendistro.elasticsearch.performanceanalyzer.reader;
 
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.metrics.AllMetrics;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.metrics.PerformanceAnalyzerMetrics;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.reader_writer_shared.Event;
 import java.sql.SQLException;
@@ -31,6 +32,11 @@ public class ClusterDetailsEventProcessorTestHelper extends AbstractReaderTests 
 
   public void addNodeDetails(String nodeId, String address, boolean isMasterNode) {
     nodeDetails.add(createNodeDetailsMetrics(nodeId, address, isMasterNode));
+  }
+
+  public static ClusterDetailsEventProcessor.NodeDetails newNodeDetails(final String nodeId, final String address,
+                                                                        final boolean isMasterNode) {
+    return createNodeDetails(nodeId, address, isMasterNode);
   }
 
   public void generateClusterDetailsEvent() {
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/stats/collectors/SampleAggregator.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/stats/collectors/SampleAggregator.java
@@ -37,7 +37,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicLong;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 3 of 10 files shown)


yojs:
Issue #, if available: 51

Description of changes: Added unit tests for rca.samplers and rca.util

Tests:

// In this section, can we mention what scenarios are we covering in the tests we add please ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/ControllerUtil.java
```diff
@@ -0,0 +1,21 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca;
+
+/**
+ * ControllerUtil is a singleton which provides visibility into {@link RcaController}'s state
+ */
+public class ControllerUtil {
+    public static final ControllerUtil API = new ControllerUtil();
```
yojs, line 7: I would suggest we don't use static classes or initializes as much as possible. 
The thing is your UT framework initializes once. If you set something in a static object in one of the tests, it stays there (unless explicitly reset after the tests pass). This can cause weird behaviors in your tests based on the order they execute etc.
If we can remove them, it will be great.

sidheart: I also want to avoid static classes / members where possible. That was the intent here. rcaEnabled previously was a static member of the RcaController that could only be set via a file read through a controller method. This causes issues for the reason you mentioned in your comment: testing with static classes/methods/members sucks. That's why I extracted the read/write logic for isRcaEnabled into the singleton class here. I did this in a few other places so that I could avoid static classes and bytecode hijacking via PowerMock.

sidheart: I removed the ControllerUtil, the RcaController can be refactored at a future time

build.gradle
```diff
@@ -230,11 +230,14 @@ dependencies {
     compile group: 'org.apache.logging.log4j', name: 'log4j-core', version: '2.13.0'
     compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.9'
     compile group: 'commons-io', name: 'commons-io', version: '2.3'
+    compile 'org.projectlombok:lombok:1.18.12'
     implementation 'io.grpc:grpc-netty-shaded:1.25.0'
     implementation 'io.grpc:grpc-protobuf:1.25.0'
     implementation 'io.grpc:grpc-stub:1.25.0'
     implementation 'javax.annotation:javax.annotation-api:1.3.2'
 
+    annotationProcessor 'org.projectlombok:lombok:1.18.12'
```
yojs, line 239: We discussed that we will remove this new dependency, right ?

sidheart: Fully removed in the latest PR

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
```diff
@@ -125,11 +132,27 @@ public static NodeDetails getCurrentNodeDetails() {
...
+    /**
+     * This constructor is provided expressly for testing purposes. In general, NodeDetails should only be
+     * retrieved through method calls such as {@link ClusterDetailsEventProcessor#getCurrentNodeDetails()}
+     * @param id The node's id
+     * @param hostAddress The host's address
+     * @param role The node's role e.g. data-node
+     * @param isMasterNode Whether or not the node is a master node
+     */
+    @VisibleForTesting
+    public NodeDetails(String id, String hostAddress, String role, Boolean isMasterNode) {
```
rguo-aws, line 149: We don't need to add a constructor specifically used by unit test. you can work around this by using a utility function API provided by another unit test. (createNodeDetailsMetrics)

sidheart: Great catch! I've improved the design as you suggested and will remember this pattern for the future.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessorTestHelper.java
```diff
@@ -33,6 +34,11 @@ public void addNodeDetails(String nodeId, String address, boolean isMasterNode)
     nodeDetails.add(createNodeDetailsMetrics(nodeId, address, isMasterNode));
   }
 
+  public static ClusterDetailsEventProcessor.NodeDetails newNodeDetails(String nodeId, String address,
```
yojs, line 37: should we make all the arguments `final` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
```diff
@@ -20,6 +20,7 @@
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.RcaControllerHelper;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.reader_writer_shared.Event;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.util.JsonConverter;
+import com.google.common.annotations.VisibleForTesting;
```
yojs, line 23: We added this, but are we using this ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
```diff
@@ -29,6 +30,7 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
+
```
yojs, line 33: Maybe we can remove this file from change list ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/RcaEnabledSampler.java
```diff
@@ -29,12 +30,12 @@ public void sample(SampleAggregator sampleCollector) {
     sampleCollector.updateStat(RcaRuntimeMetrics.RCA_ENABLED, "", isRcaEnabled());
   }
 
-  private int isRcaEnabled() {
+  @VisibleForTesting
+  int isRcaEnabled() {
```
yojs, line 34: I would expect this to return a `boolean` :)

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessorTestHelper.java
sidheart: Done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
sidheart: Removed

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
sidheart: Removed

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/RcaEnabledSampler.java
sidheart: Done

PR #138, opened 2020-04-07
Added unit tests for the rca.net.handler package

sidheart:
*Issue #, if available:* 51

*Description of changes:* Added unit tests for the rca.net.handler package

*Tests:*


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/net/CompositeSubscribeRequest.java
```diff
@@ -56,4 +58,17 @@ public SubscribeMessage getSubscribeMessage() {
   public StreamObserver<SubscribeResponse> getSubscribeResponseStream() {
     return subscribeResponseStream;
   }
+
+  @Override
+  public boolean equals(final Object o) {
```
yojs, line 63: We should refrain from adding equals just for the purpose of testing. We should rather use a helper if Tests are the only ones that need it.

sidheart:
This is now largely covered by the new rca.net unit tests

PR #141, opened 2020-04-09
Adding code changes for Hot Shards RCA.

khushbr:
The RCA currrntly identifies a hot shard within an index based on following dimensions:
1. CPU Utilization
2. IO Total SysCallRate
3. IO Total Throughput

### DIFF
--- a/pa_config/rca_idle_master.conf
+++ b/pa_config/rca_idle_master.conf
@@ -1,5 +1,5 @@
 {
-  "analysis-graph-implementor": "com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.DummyGraph",
+  "analysis-graph-implementor": "com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.ElastiSearchAnalysisGraph",
   // it can be file:// or s3://
   "rca-store-location": "s3://sifi-store/rcas/",
 
--- a/pa_config/rca_idle_master.conf
+++ b/pa_config/rca_idle_master.conf
@@ -1,5 +1,5 @@
 {
-  "analysis-graph-implementor": "com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.ElastiSearchAnalysisGraph",
+  "analysis-graph-implementor": "com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.ElasticSearchAnalysisGraph",
   // it can be file:// or s3://
   "rca-store-location": "s3://sifi-store/rcas/",
 
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/HotNodeSummary.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/HotNodeSummary.java
@@ -48,11 +48,20 @@ public class HotNodeSummary extends GenericSummary {
   private static final Logger LOG = LogManager.getLogger(HotNodeSummary.class);
   private final String nodeID;
   private final String hostAddress;
+  private final List<HotShardSummary> hotShardSummaryList;
 
   public HotNodeSummary(String nodeID, String hostAddress) {
     super();
     this.nodeID = nodeID;
     this.hostAddress = hostAddress;
+    this.hotShardSummaryList = new ArrayList<>();
+  }
+
+  public HotNodeSummary(String nodeID, String hostAddress, List<HotShardSummary> hotShardSummaryList) {
+    super();
+    this.nodeID = nodeID;
+    this.hostAddress = hostAddress;
+    this.hotShardSummaryList = hotShardSummaryList;
   }
 
   public String getNodeID() {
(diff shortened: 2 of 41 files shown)


pa_config/rca_idle_master.conf
```diff
@@ -1,5 +1,5 @@
 {
-  "analysis-graph-implementor": "com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.DummyGraph",
+  "analysis-graph-implementor": "com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.ElastiSearchAnalysisGraph",
```
rguo-aws, line 2: I think there is a typo here. is it supposed to be "ElasticSearchAnalysisGraph" ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElastiSearchAnalysisGraph.java
```diff
@@ -88,5 +94,32 @@ public void construct() {
...
+    // High CPU Usage RCA
+    HighCPUShardRca highCPUShardRca = new HighCPUShardRca(5, 12, cpuUsage, ioTotThroughput, ioTotSyscallRate);
+    highCPUShardRca.addTag(TAG_LOCUS, LOCUS_DATA_MASTER_NODE);
+    highCPUShardRca.addAllUpstreams(Arrays.asList(cpuUsage, ioTotThroughput, ioTotSyscallRate));
+
+    // Hot Shard RCA which consumes the HighCPUShardRca
+    HotShardClusterRca hotShardClusterRca = new HotShardClusterRca(12, highCPUShardRca);
+    hotShardClusterRca.addTag(TAG_LOCUS, LOCUS_MASTER_NODE);
+    hotShardClusterRca.addAllUpstreams(Collections.singletonList(highCPUShardRca));
+    hotShardClusterRca.addTag(RcaTagConstants.TAG_AGGREGATE_UPSTREAM, LOCUS_DATA_NODE);
```
rguo-aws, line 123: do we still need to set TAG_AGGREGATE_UPSTREAM on master node ? I assume HighCPUShardRca will only run on data node right ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/HotNodeSummary.java
```diff
@@ -48,11 +48,20 @@
   private static final Logger LOG = LogManager.getLogger(HotNodeSummary.class);
   private final String nodeID;
   private final String hostAddress;
+  private final List<HotShardSummary> hotShardSummaryList;
```
rguo-aws, line 51: can we remove this private list and add all HotShardSummary into its GenericSumary list ? The idea is HotNodeSummry is a generic type of summary which does not need to know what type of summary it has in its nested summary list.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/hotshard/IndexShardKey.java
```diff
@@ -0,0 +1,54 @@
...
+ * express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.hotshard;
+
+public class IndexShardKey {
+
+    private final String indexName;
+    private final String shardId;
```
rguo-aws, line 21: can we declare shard id as int ?

pa_config/rca_idle_master.conf
khushbr: Right, fixing this typo

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/HotNodeSummary.java
khushbr: As discussed, using the generic list would mean we use the `instanceof` to decide the list type, which is something we want to avoid.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/hotshard/IndexShardKey.java
khushbr: I refrained from using String as within the hashtable, we are using a composite key of index name, shard Id and Node Name and having a string avoids the multiple conversions we will have to take with an int.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElastiSearchAnalysisGraph.java
khushbr: As discussed, we will be needing the `TAG_AGGREGATE_UPSTREAM` here.

PR #144, opened 2020-04-17
Code for integrating the muting API with AnalysisGraph

khushbr:
*Description of changes:*  **Code for integrating the muting API with AnalysisGraph.**

The RcaController thread will check the muted_rcas.conf file and update the Analysis Graph for muted RCAs. Plugged into existing code to disable reading metrics if node is Muted.


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -255,11 +268,42 @@ private void readRcaEnabledFromConf() {
         });
   }
 
+  /**
+   * Reads the mutedRCAs value from the conf file.
+   */
+  private void readMutedRcasFromConf() {
+    // all the config filed are located in same location as RCA_ENABLED_CONF_LOCATION
```
yojs, line 275: typo: files ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -85,6 +88,11 @@
   // This needs to be volatile as the RcaConfPoller writes it but the Nanny reads it.
   private static volatile boolean rcaEnabled = false;
 
+  private String mutedRcasDefaultValue = "";
```
yojs, line 91: Maybe `final` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -85,6 +88,11 @@
...
+
+  // This needs to be volatile as the RcaConfPoller writes it but the Nanny reads it.
+  private static volatile String mutedRcas = "";
```
yojs, line 94: I am thinking instead of storing a list of rcas in a comma separated string, we should probably process it and it should be an array of Strings ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -255,11 +268,42 @@ private void readRcaEnabledFromConf() {
         });
   }
 
+  /**
+   * Reads the mutedRCAs value from the conf file.
```
yojs, line 272: we should also add to the doc comments, what is a valid format for this file ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -216,7 +224,12 @@ public void run() {
           }
         }
 
+        // Read the Muted RCAs value
+        readMutedRcasFromConf();
```
yojs, line 228: We can read and update the muted RCAs from conf here at once ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -255,11 +268,42 @@ private void readRcaEnabledFromConf() {
...
+   * Reads the mutedRCAs value from the conf file.
+   */
+  private void readMutedRcasFromConf() {
+    // all the config filed are located in same location as RCA_ENABLED_CONF_LOCATION
+    Path filePath = Paths.get(RCA_ENABLED_CONF_LOCATION, MUTED_RCAS_CONF_FILE);
+
+    Util.invokePrivileged(
+            () -> {
+              try (Scanner sc = new Scanner(filePath)) {
+                mutedRcas = sc.nextLine();
```
yojs, line 281: Once we read the RCAs, we should also proess the string and make sure that all the RCAs mentioned are valid graph nodes. If they are not, then we should write an error log and discard that element.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Yup,  a typo. Updating this.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Agreed, makes more sense to have it as `final`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Agreed. 
Initially I had planned to use an array of strings for mutedRCA cluster settings but ES cluster settings only support list type with special handling. Since we are no longer relying on cluster settings, I will revert this to use an array of strings.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Right, adding this information.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Done. Merged the read and muting the rcas in analysis graph into single function.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Agreed, made the change in latest revision

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -273,6 +317,9 @@ private void updateRcaState() {
           restart();
           PerformanceAnalyzerApp.RCA_RUNTIME_METRICS_AGGREGATOR.updateStat(
               RcaRuntimeMetrics.RCA_RESTARTED_BY_OPERATOR, "", 1);
+
+          // Update Analysis Graph for Muted RCAs
+          updateAnalysisGraphWithMutedRcas();
```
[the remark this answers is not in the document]
khushbr: Thank You for pointing this out. Have fixed in latest revision.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -77,6 +78,8 @@
 
   private static final String RCA_ENABLED_CONF_FILE = "rca_enabled.conf";
 
+  private static final String MUTED_RCAS_CONF_FILE = "muted_rcas.conf";
```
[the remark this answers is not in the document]
khushbr: Re-using the existing `rca.conf` file(s) now.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -77,6 +79,8 @@
...
```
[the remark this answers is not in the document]
khushbr: We are now using the util function `RcaControllerHelper.pickRcaConfForRole()` for picking the conf file, the function has a doc comment explaining the function of file. 

This is not being used and I will remove it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -216,7 +223,14 @@ public void run() {
           }
         }
 
+        // Update the RCA state
```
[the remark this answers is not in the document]
khushbr: Right, we can remove this.

PR #154, opened 2020-04-21
Ignore the rca.stats.eval.impl.vals package from coverage

sidheart:
This package only contains constructors and getters

*Issue #, if available:* 51 

*Description of changes:* Ignore the rca.stats.eval.impl.vals package from coverage

*Tests:*


sidheart:
Will attempt to cover these items in a higher level test, or will restore the previous coverage PR

PR #155, opened 2020-04-22
Code for integrating the muting API with AnalysisGraph

khushbr:
*Description of changes:* 
The RcaController thread will check the muted_rcas.conf file and update the Analysis Graph for muted RCAs. Plugged into existing code to disable reading metrics if node is Muted.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
@@ -56,8 +56,10 @@ import java.lang.reflect.InvocationTargetException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.sql.SQLException;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Scanner;
+import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -77,6 +79,8 @@ public class RcaController {
 
   private static final String RCA_ENABLED_CONF_FILE = "rca_enabled.conf";
 
+  private static final String MUTED_RCAS_CONF_FILE = "muted_rcas.conf";
+
   private final ScheduledExecutorService netOpsExecutorService;
   private final boolean useHttps;
 
@@ -84,6 +88,9 @@ public class RcaController {
 
   // This needs to be volatile as the RcaConfPoller writes it but the Nanny reads it.
   private static volatile boolean rcaEnabled = false;
+
+  // This needs to be volatile as the RcaConfPoller writes it but the Nanny reads it.
+  private static volatile long lastModifiedInMemory = 0;
 
   // This needs to be volatile as the NodeRolePoller writes it but the Nanny reads it.
   private volatile NodeRole currentRole = NodeRole.UNKNOWN;
@@ -216,7 +223,14 @@ public class RcaController {
           }
         }
 
+        // Update the RCA state
         updateRcaState();
+
+        // Update Analysis graph with Muted RCAs value
+        if (rcaEnabled && rcaScheduler != null && rcaScheduler.getState() == RcaSchedulerState.STATE_STARTED) {
+          readAndUpdateMutesRcas();
+        }
+
         long duration = System.currentTimeMillis() - startTime;
         if (duration < rcaStateCheckIntervalMillis) {
           Thread.sleep(rcaStateCheckIntervalMillis - duration);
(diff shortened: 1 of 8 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -255,6 +269,40 @@ private void readRcaEnabledFromConf() {
         });
   }
 
+  /**
+   * Reads the mutedRCAList value from the rca.conf file, performs validation on the param value
+   * provided and on successful validation, updates the AnalysisGraph with muted RCA value.
+   *
+   */
+  private void readAndUpdateMutesRcas() {
```
yojs, line 277: I am thinking if we should move this [here](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/blob/3267f003f4b5acc70f8c603472cc1ed44859a310/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/scheduler/RCASchedulerTask.java#L374) ? `RCASchedulerTask` is the component that actually deals with the Graph.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -85,6 +89,9 @@
   // This needs to be volatile as the RcaConfPoller writes it but the Nanny reads it.
   private static volatile boolean rcaEnabled = false;
 
+  // This needs to be volatile as the RcaConfPoller writes it but the Nanny reads it.
+  private static volatile long lastModifiedInMemory = 0;
```
yojs, line 93: This is in milliseconds right? We should add that to the name ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/ConfJsonWrapper.java
```diff
@@ -103,5 +111,6 @@ public void setDatastoreRcaLogDirectory(String rcaLogLocation) {
     this.analysisGraphEntryPoint = analysisGraphEntryPoint;
     this.networkQueueLength = networkQueueLength;
     this.perVertexBufferLength = perVertexBufferLength;
+    this.mutedRcaList = Arrays.asList(mutedRcas.split("\\s*,\\s*"));
```
yojs, line 114: can we add a line of comment here ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/RcaConf.java
```diff
@@ -84,6 +89,10 @@ public String getConfigFileLoc() {
     return configFileLoc;
   }
 
+  public long getLastModifiedTime() {
```
yojs, line 92: Let's add a comment as this is the last modified time of the rca.conf file itself

src/test/resources/rca/rca.conf
```diff
@@ -41,5 +41,7 @@
     // How often the sqlite file be repeated in seconds. This file contains RCAs and therefore rotating it too frequently
     // might not be as fruitful as there might not be any data.
     "rotation-period-seconds": 21600
-  }
+  },
+
+  "muted-rcas": "rca1, rca2"
```
yojs, line 46: Can we also have a couple of negative test cases and one where the comma separated rcas don't have spaces between them ?

yojs: We should add a test at the graph creation level where no Graph node matches the names of the Rca being asked to be muted and And one where a graph node is successfully muted

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Right, Will update the variable name to reflect this.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/ConfJsonWrapper.java
khushbr: Will add a comment line explaining what the regex is doing.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/RcaConf.java
khushbr: Makes sense. Will add.

src/test/resources/rca/rca.conf
khushbr: Agreed, the test I am writing currently and will cover the happy cases and otherwise.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
yojs: The other place would be [here](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/blob/3267f003f4b5acc70f8c603472cc1ed44859a310/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java#L210). After we check if the RcaEnabled config change, we check if the rca.conf has been updated (by checking the last modified time). If so, then we check if the muted Rca list has been updated and if so, we do validate and update the muted RCA list. If it hasn't we move on, until next time. What do you think ?

PR #167, opened 2020-04-25
Adding final changes for integrating Muting API with scheduler

khushbr:
*Issue #, if available:*

*Description of changes:* The RcaController thread will check the muted_rcas.conf file and update the Analysis Graph for muted RCAs. Plugged into existing code to disable reading metrics if node is Muted.
- Added UTs
- Incoporated PR Feedback

The API behavior : The API will be invoked with full list of RCAs that are expected to be muted. The input will be provided as a comma separated string, similar to `"CPU_Utilization, Heap_AllocRate"`. The API implementation takes care of validating the RCA names provided in the input and any leading/trailing spaces.
1. If API is invoked with at least 1 valid RCA, the muted RCA Graph is updated to retain only this valid RCA(s).
1. If API is invoked with an empty string, the muted RCA Graph is updated and all previous existing RCAs in the graph are removed.
2. If API is invoked with an all invalid RCAs, no action is taken. 

*Tests:* UT, Dev Stack Testing

### DIFF
--- a/pa_config/rca.conf
+++ b/pa_config/rca.conf
@@ -41,5 +41,7 @@
   // RCA settings
   "high-heap-usage-old-gen-rca": {
     "top-k" : 3
-  }
+  },
+
+  "muted-rcas": ""
 }
--- a/pa_config/rca.conf
+++ b/pa_config/rca.conf
@@ -36,5 +36,10 @@
     // How often the sqlite file be repeated in seconds. This file contains RCAs and therefore rotating it too frequently
     // might not be as fruitful as there might not be any data.
     "rotation-period-seconds": 21600
+  },
+
+  // RCA settings
+  "high-heap-usage-old-gen-rca": {
+    "top-k" : 3
   }
 }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
@@ -56,12 +56,15 @@ import java.lang.reflect.InvocationTargetException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.sql.SQLException;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Scanner;
+import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -85,6 +88,9 @@ public class RcaController {
   // This needs to be volatile as the RcaConfPoller writes it but the Nanny reads it.
   private static volatile boolean rcaEnabled = false;
 
+  // This needs to be volatile as the RcaConfPoller writes it but the Nanny reads it.
+  private static volatile long lastModifiedTimeInMillisInMemory = 0;
+
   // This needs to be volatile as the NodeRolePoller writes it but the Nanny reads it.
   private volatile NodeRole currentRole = NodeRole.UNKNOWN;
 
@@ -99,6 +105,8 @@ public class RcaController {
   private QueryRcaRequestHandler queryRcaRequestHandler;
 
   private SubscriptionManager subscriptionManager;
+
+  private RcaConf rcaConf;
 
   private final String RCA_ENABLED_CONF_LOCATION;
   private final long rcaStateCheckIntervalMillis;
@@ -135,7 +143,7 @@ public class RcaController {
   }
 
   private void start() {
-    final RcaConf rcaConf = RcaControllerHelper.pickRcaConfForRole(currentRole);
+    rcaConf = RcaControllerHelper.pickRcaConfForRole(currentRole);
     try {
       subscriptionManager.setCurrentLocus(rcaConf.getTagMap().get("locus"));
       List<ConnectedComponent> connectedComponents = RcaUtil.getAnalysisGraphComponents(rcaConf);
@@ -215,8 +223,8 @@ public class RcaController {
             checkUpdateNodeRole(nodeDetails);
           }
         }
-
         updateRcaState();
+
         long duration = System.currentTimeMillis() - startTime;
         if (duration < rcaStateCheckIntervalMillis) {
           Thread.sleep(rcaStateCheckIntervalMillis - duration);
@@ -253,6 +261,51 @@ public class RcaController {
             rcaEnabled = rcaEnabledDefaultValue;
           }
         });
+
+    // If RCA is enabled, update Analysis graph with Muted RCAs value
+    if (rcaEnabled) {
+      LOG.debug("Updating Analysis Graph with Muted RCAs");
+      readAndUpdateMutesRcas();
+    }
+  }
+
+  /**
+   * Reads the mutedRCAList value from the rca.conf file, performs validation on the param value
+   * provided and on successful validation, updates the AnalysisGraph with muted RCA value.
+   *
+   * <p>In case all the RCAs in param value are incorrect, return without any update.
+   */
+  private void readAndUpdateMutesRcas() {
+    // If the rca config file has been updated since the lastModifiedTimeInMillisInMemory in memory,
+    // refresh the `muted-rcas` value from rca config file.
+    long lastModifiedTimeInMillisOnDisk = rcaConf.getLastModifiedTime();
+    if (lastModifiedTimeInMillisOnDisk > lastModifiedTimeInMillisInMemory) {
+      try {
+        Set<String> rcasForMute = new HashSet<>(rcaConf.getMutedRcaList());
+        LOG.info("RCAs provided for muting : {}", rcasForMute);
+
+        Set<String> graphNodeNames = new HashSet<>();
+        RcaUtil.getAnalysisGraphComponents(rcaConf).forEach(
+                connectedComponent -> connectedComponent.addNodeNames(graphNodeNames));
+
+        // Update rcasForMute to retain only valid RCAs
+        rcasForMute.retainAll(graphNodeNames);
+
+        // If rcasForMute post validation is empty but rcaConf.getMutedRcaList() is not empty
+        // all the input RCAs are incorrect, return.
+        if (rcasForMute.isEmpty() && !rcaConf.getMutedRcaList().isEmpty()) {
+          LOG.error("Incorrect RCA(s): {}, cannot be muted. Valid RCAs: {}",
+                  rcaConf.getMutedRcaList(), graphNodeNames);
+          return;
+        }
+
+        LOG.info("Updating the muted RCA Graph to : {}", rcasForMute);
+        Stats.getInstance().updateMutedGraphNodes(rcasForMute);
+      } catch (Exception e) {
+        LOG.error("Couldn't read/update the muted RCAs.", e);
+      }
+    }
+    lastModifiedTimeInMillisInMemory = lastModifiedTimeInMillisOnDisk;
   }
 
   /**
@@ -286,7 +339,6 @@ public class RcaController {
       }
     }
   }
-
 
   private void removeRcaRequestHandler() {
     try {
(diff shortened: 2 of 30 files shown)


pa_config/rca.conf
```diff
@@ -36,5 +36,7 @@
     // How often the sqlite file be repeated in seconds. This file contains RCAs and therefore rotating it too frequently
     // might not be as fruitful as there might not be any data.
     "rotation-period-seconds": 21600
-  }
+  },
+
+  "muted-rcas": "rca1, rca2"
```
yojs, line 41: We should remove this, right ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -215,8 +222,13 @@ public void run() {
             checkUpdateNodeRole(nodeDetails);
           }
         }
-
         updateRcaState();
+
+        // Update Analysis graph with Muted RCAs value
+        if (rcaEnabled && rcaScheduler != null && rcaScheduler.getState() == RcaSchedulerState.STATE_STARTED) {
```
yojs, line 228: As we discussed, we want to update the mutedRcas before the call to `updateRcaState()`

yojs: The mutedRcas is an update to a static datastructure. It should work, whether Rca is enabled or not. But to avoid processing, we can just check for the `rcaEnabled` flag ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -255,6 +267,47 @@ private void readRcaEnabledFromConf() {
...
+  /**
+   * Reads the mutedRCAList value from the rca.conf file, performs validation on the param value
+   * provided and on successful validation, updates the AnalysisGraph with muted RCA value.
+   *
+   * <p>In case all the RCAs in param value are incorrect, return without any update.
+   */
+  private void readAndUpdateMutesRcas() {
+    // If the rca config file has been updated since the lastModifiedTimeInMillisInMemory in memory,
+    // refresh the `muted-rcas` value from rca config file.
+    final RcaConf rcaConf = RcaControllerHelper.pickRcaConfForRole(currentRole);
```
yojs, line 279: The `start()` method already calls `pickRcaConfForRole`. As JSON processing is expensive and this also means that in theory `start()` and `readAndUpdateMutesRcas` can work on different `rca.conf` if it were to be updated between the two calls. 

Long story short, can we store `RcaConf` as a class member after reading it once ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
yojs: We should have a test to make sure that mutedRcas is always accessible

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -255,6 +267,47 @@ private void readRcaEnabledFromConf() {
...
+                connectedComponent -> graphNodeNames.addAll(connectedComponent.getNodeNames()));
+
+        Set<String> validRcasForMute = rcasForMute.stream()
+                .filter(rcaForMute -> graphNodeNames.contains(rcaForMute))
+                .collect(Collectors.toSet());
+
+        // If validRcasForMute is empty but rcasForMute is not empty
+        // all the input RCAs are incorrect, return.
+        if (validRcasForMute.isEmpty() && !rcasForMute.isEmpty()) {
+          LOG.info("Incorrect RCA value, cannot be muted : {}", rcasForMute);
```
yojs, line 297: `error` log ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -255,6 +267,47 @@ private void readRcaEnabledFromConf() {
...
+          return;
+        }
+
+        LOG.info("Updating the muted RCA Graph to : {}", validRcasForMute);
+        Stats.getInstance().updateMutedGraphNodes(validRcasForMute);
+      } catch (Exception e) {
+        LOG.error("Couldn't read/update the muted RCAs. Ran into {}", e.getMessage());
+        e.printStackTrace();
```
yojs, line 305: Instead of this, we should add this to the log ?
LOG.error("Couldn't read/update the muted RCAs.", e);

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
yojs: Thinking more about it, when people make the API call from the writer, they wouldn't even know as part of the API that they are referring to a non-existent RCA or other error. It will be silently emitted in the logs but the API will return OK status.

Because the Reader has the knowledge of the list of RCAs, it might make sense for the API on the writer to just proxy this to the reader and the READER do a validation and return error as the response itself and not just as silent logs. Thoughts ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -255,6 +267,47 @@ private void readRcaEnabledFromConf() {
...
+    // refresh the `muted-rcas` value from rca config file.
+    final RcaConf rcaConf = RcaControllerHelper.pickRcaConfForRole(currentRole);
+    long lastModifiedTimeInMillisOnDisk = rcaConf.getLastModifiedTime();
+    if (lastModifiedTimeInMillisOnDisk > lastModifiedTimeInMillisInMemory) {
+      try {
+        List<String> rcasForMute = rcaConf.getMutedRcaList();
+        LOG.info("RCAs provided for muting : {}", rcasForMute);
+
+        Set<String> graphNodeNames = new HashSet<>();
+        RcaUtil.getAnalysisGraphComponents(rcaConf).forEach(
```
yojs, line 287: We are iterating through the list of nodes three times: 1. in `getAllNodes()`, then again in `getNodeNames()` and then here. And we are create new HasSets everytime. Let's try to minimize the HashSet creation and iterations.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaControllerTest.java
```diff
@@ -147,11 +152,77 @@ public void readRcaEnabledFromConf() throws IOException {
...
+    readAndUpdateMutesRcas.invoke(rcaController);
+  }*/
+
   @Test
   public void nodeRoleChange() throws IOException {
     changeRcaRunState(RcaState.RUN);
     masterIP = "10.10.192.168";
     setMyIp(masterIP, AllMetrics.NodeRole.ELECTED_MASTER);
+
+    // lastModifiedTimeInMillisInMemory < lastModifiedTimeInMillisOnDisk
```
yojs, line 225: remove ?

src/test/resources/rca/rca.conf
```diff
@@ -41,5 +41,7 @@
     // How often the sqlite file be repeated in seconds. This file contains RCAs and therefore rotating it too frequently
     // might not be as fruitful as there might not be any data.
     "rotation-period-seconds": 21600
-  }
+  },
+
+    "muted-rcas": "CPU_Utilization, Heap_AllocRate"
```
yojs, line 46: Maybe, we should not add muted RCA nodes to these configs that are generically imported by all tests. This will mute some RCA nodes underneath. As muted RCA is a special config, we should exclusively create rca.conf for these test cases. Thoughts ?

pa_config/rca.conf
khushbr: Right. Have replaced this with an empty string.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: > The mutedRcas is an update to a static datastructure. It should work, whether Rca is enabled or not. But to avoid processing, we can just check for the `rcaEnabled` flag ?

Agreed, Have updated to move this under [readRcaEnabledFromConf](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/blob/master/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java#L242)

khushbr: > We should have a test to make sure that mutedRcas is always accessible

Right. I am adding one.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Agreed, makes sense.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Yup, updated this.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaControllerTest.java
khushbr: Done.

src/test/resources/rca/rca.conf
khushbr: I was debating the need for having a separate config file for testing muted RCAs to ensure the orthogonality. Have added one.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: We can discuss this further since this behavior would involve more elaborate changes through the code. Currently, the expectation is for the operator to either perform a GET after PUT to verify the muted API result or check the logs for errors. 

I agree this is something we should ideally do.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Done, have updated the code to use a single hashmap, rather than creating multiple.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -253,6 +261,51 @@ private void readRcaEnabledFromConf() {
...
+    // If the rca config file has been updated since the lastModifiedTimeInMillisInMemory in memory,
+    // refresh the `muted-rcas` value from rca config file.
+    long lastModifiedTimeInMillisOnDisk = rcaConf.getLastModifiedTime();
+    if (lastModifiedTimeInMillisOnDisk > lastModifiedTimeInMillisInMemory) {
+      try {
+        Set<String> rcasForMute = new HashSet<>(rcaConf.getMutedRcaList());
+        LOG.info("RCAs provided for muting : {}", rcasForMute);
+
+        Set<String> graphNodeNames = new HashSet<>();
+        RcaUtil.getAnalysisGraphComponents(rcaConf).forEach(
```
yojs, line 288: This is not the right method to call. This will re-initialize the entire Analysis graph.

yojs: I was thinking we can do something much simpler. As [here](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/pull/167/commits/54bc23619d9c66643fcea6afb74fa2f188ab4a94#diff-c1b8be1a0f84a6a0b5f51bb1041f9c9bL34)  we are iterating over all nodes, we also create another HashSet<String> that contains the node Names only. Then the method [here](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/pull/167/commits/54bc23619d9c66643fcea6afb74fa2f188ab4a94#diff-c1b8be1a0f84a6a0b5f51bb1041f9c9bR116) serves as a getter for it. This avoids traversing all the nodes twice and also re-creating the HashSet on each call.

PR #168, opened 2020-04-28
Muting API

khushbr:
Description of changes: The RcaController thread will check the muted_rcas.conf file and update the Analysis Graph for muted RCAs. Plugged into existing code to disable reading metrics if node is Muted.

    Added UTs
    Incoporated PR Feedback

The API behavior : The API will be invoked with full list of RCAs that are expected to be muted. The input will be provided as a comma separated string, similar to "CPU_Utilization, Heap_AllocRate". The API implementation takes care of validating the RCA names provided in the input and any leading/trailing spaces.

    If API is invoked with at least 1 valid RCA, the muted RCA Graph is updated to retain only this valid RCA(s).
    If API is invoked with an empty string, the muted RCA Graph is updated and all previous existing RCAs in the graph are removed.
    If API is invoked with an all invalid RCAs, no action is taken.

Tests: UT, Dev Stack Testing

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
@@ -286,7 +286,7 @@ public class RcaController {
 
         Set<String> graphNodeNames = new HashSet<>();
         RcaUtil.getAnalysisGraphComponents(rcaConf).forEach(
-                connectedComponent -> connectedComponent.addNodeNames(graphNodeNames));
+                connectedComponent -> graphNodeNames.addAll(connectedComponent.getNodeNames()));
 
         // Update rcasForMute to retain only valid RCAs
         rcasForMute.retainAll(graphNodeNames);
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
@@ -284,18 +284,14 @@ public class RcaController {
         Set<String> rcasForMute = new HashSet<>(rcaConf.getMutedRcaList());
         LOG.info("RCAs provided for muting : {}", rcasForMute);
 
-        Set<String> graphNodeNames = new HashSet<>();
-        RcaUtil.getAnalysisGraphComponents(rcaConf).forEach(
-                connectedComponent -> graphNodeNames.addAll(connectedComponent.getNodeNames()));
-
         // Update rcasForMute to retain only valid RCAs
-        rcasForMute.retainAll(graphNodeNames);
+        rcasForMute.retainAll(ConnectedComponent.getNodeNames());
 
         // If rcasForMute post validation is empty but rcaConf.getMutedRcaList() is not empty
         // all the input RCAs are incorrect, return.
         if (rcasForMute.isEmpty() && !rcaConf.getMutedRcaList().isEmpty()) {
-          LOG.error("Incorrect RCA(s): {}, cannot be muted. Valid RCAs: {}",
-                  rcaConf.getMutedRcaList(), graphNodeNames);
+          LOG.error("Incorrect RCA(s): {}, cannot be muted. Valid RCAs: {}, Muted RCAs: {}",
+                  rcaConf.getMutedRcaList(), ConnectedComponent.getNodeNames(), Stats.getInstance().getMutedGraphNodes());
           return;
         }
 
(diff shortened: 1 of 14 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/ConnectedComponent.java
```diff
@@ -114,4 +117,10 @@ public void addLeafNode(Node node) {
     }
     return dependencyOrderedNodes;
   }
+
+  public Set<String> getNodeNames() {
+    getAllNodes();
```
yojs, line 122: `getAllNodes()` is an expensive call. It should only be called if `nodeNames` is uninitialized.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -253,6 +261,51 @@ private void readRcaEnabledFromConf() {
...
+    // refresh the `muted-rcas` value from rca config file.
+    long lastModifiedTimeInMillisOnDisk = rcaConf.getLastModifiedTime();
+    if (lastModifiedTimeInMillisOnDisk > lastModifiedTimeInMillisInMemory) {
+      try {
+        Set<String> rcasForMute = new HashSet<>(rcaConf.getMutedRcaList());
+        LOG.info("RCAs provided for muting : {}", rcasForMute);
+
+        Set<String> graphNodeNames = new HashSet<>();
+        RcaUtil.getAnalysisGraphComponents(rcaConf).forEach(
+                connectedComponent -> graphNodeNames.addAll(connectedComponent.getNodeNames()));
```
yojs, line 289: I don't get it, why can't we just call `getNodeNames()` instead of calling the expensive method `getAnalysisGraphComponents` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -253,6 +261,51 @@ private void readRcaEnabledFromConf() {
...
+
+        // Update rcasForMute to retain only valid RCAs
+        rcasForMute.retainAll(graphNodeNames);
```
yojs, line 292: Ideally this one should be:
`rcasForMute.retainAll(getNodeNames());`, right ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -253,6 +261,51 @@ private void readRcaEnabledFromConf() {
...
+
+        // If rcasForMute post validation is empty but rcaConf.getMutedRcaList() is not empty
+        // all the input RCAs are incorrect, return.
+        if (rcasForMute.isEmpty() && !rcaConf.getMutedRcaList().isEmpty()) {
+          LOG.error("Incorrect RCA(s): {}, cannot be muted. Valid RCAs: {}",
```
ktkrg, line 297: nit: suggest renaming this(and other instances) to node muting instead of rca muting. We're not restricting this to just rca nodes anywhere.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/ConnectedComponent.java
khushbr: With current implementation, we are returning `nodeNames` without making the null check. If the `nodeNames` is not populated, then it should fail.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Agreed, so now we are maintaining a set which gets populated at the time of graph creation and is accessible via `getNodeNames()`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: Done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/ConnectedComponent.java
```diff
@@ -29,12 +29,11 @@
 
   /* The elements in the inner list can be executed in parallel. Two inner lists have to be executed in order. */
   private List<List<Node<?>>> dependencyOrderedNodes;
-  private Set<String> nodeNames;
+  private static Set<String> nodeNames = new HashSet<>();
   private int graphId;
 
   public Set<Node<?>> getAllNodes() {
     Set<Node<?>> traversed = new HashSet<>();
-    nodeNames = new HashSet<>();
```
yojs: I don't see where we are filling the set up ?

PR #173, opened 2020-04-29
Adding NPE Fix during RCA controller start up

khushbr:
*Issue #, if available:* https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/issues/172

*Description of changes:*
1. Added the rcaConf null check under `readAndUpdateMutesRcas()`
2. Moved the invocation of `readAndUpdateMutesRcas()` to after rcaConf initialization and rcaScheduler restart in `updateRcaState()`

*Tests:* Added a UT to check the NPE condition for rcaConf

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
@@ -225,6 +225,12 @@ public class RcaController {
         }
         updateRcaState();
 
+        // If RCA is enabled, update Analysis graph with Muted RCAs value
+        if (rcaEnabled) {
+          LOG.debug("Updating Analysis Graph with Muted RCAs");
+          readAndUpdateMutesRcas();
+        }
+
         long duration = System.currentTimeMillis() - startTime;
         if (duration < rcaStateCheckIntervalMillis) {
           Thread.sleep(rcaStateCheckIntervalMillis - duration);
@@ -261,12 +267,6 @@ public class RcaController {
             rcaEnabled = rcaEnabledDefaultValue;
           }
         });
-
-    // If RCA is enabled, update Analysis graph with Muted RCAs value
-    if (rcaEnabled) {
-      LOG.debug("Updating Analysis Graph with Muted RCAs");
-      readAndUpdateMutesRcas();
-    }
   }
 
   /**
@@ -276,6 +276,11 @@ public class RcaController {
    * <p>In case all the RCAs in param value are incorrect, return without any update.
    */
   private void readAndUpdateMutesRcas() {
+    if (rcaConf == null) {
+      LOG.error("RCA Conf: {} null, returning.", rcaConf);
+      return;
+    }
+
     // If the rca config file has been updated since the lastModifiedTimeInMillisInMemory in memory,
     // refresh the `muted-rcas` value from rca config file.
     long lastModifiedTimeInMillisOnDisk = rcaConf.getLastModifiedTime();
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -276,6 +276,11 @@ private void readRcaEnabledFromConf() {
    * <p>In case all the RCAs in param value are incorrect, return without any update.
    */
   private void readAndUpdateMutesRcas() {
+    if (rcaConf == null) {
```
yojs, line 279: In what case can this be `null` ? This shouldn't ever be null. RCA system depends on this config. Scheduler etc. cannot even start when this is null.

khushbr: The [rca config is initialized](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/blob/master/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java#L146) in the `start()` as part of [`updateRcaState()`](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/blob/master/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java#L334) invocation. 

The readAndUpdateMutesRcas() is invoked from [`readRcaEnabledFromConf()`](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/blob/master/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java#L250), which happens before the `updateRcaState()`call, thus leading to NPE.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -225,6 +225,12 @@ public void run() {
         }
         updateRcaState();
 
+        // If RCA is enabled, update Analysis graph with Muted RCAs value
+        if (rcaEnabled) {
+          LOG.debug("Updating Analysis Graph with Muted RCAs");
+          readAndUpdateMutesRcas();
```
yojs, line 231: This call has to be before `updateRcaState` as `updateRcaState` starts the scheduler. The scheduler is started, then it might start its run anytime. Therefore, in theory, it might execute the muted RCA node before coming around to it. Therefore, my suggestion, is move this call before `updateRcaState` this will also require adjusting the call to `rcaConf = RcaControllerHelper.pickRcaConfForRole(currentRole);`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -297,11 +297,11 @@ private void readAndUpdateMutesRcas() {
 
         LOG.info("Updating the muted RCA Graph to : {}", rcasForMute);
         Stats.getInstance().updateMutedGraphNodes(rcasForMute);
-      } catch (Exception e) {
-        LOG.error("Couldn't read/update the muted RCAs.", e);
       }
+      lastModifiedTimeInMillisInMemory = lastModifiedTimeInMillisOnDisk;
+    } catch (Exception e) {
+        LOG.error("Couldn't read/update the muted RCAs.", e);
```
yojs, line 303: You might want to emit a metric here. If we really expect never for this to happen, then we should emit a domain level metric and add it to the dashboard.

Issue #150, opened 2020-04-21
Add heap_allocation rate as a temperature dimension


yojs:
The temperature profile as it is [here](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/issues/143) is only based on CPU utilization.

ktkrg:
Closed in favor of #170

ktkrg:
Closed the wrong thing :)

PR #176, opened 2020-04-29
Temperature bug fix

ktkrg:
*Issue #, if available:*
#150 
*Description of changes:*
Adding data to a repeated element needs to be done using add* method instead of set* if we are adding data incrementally.

*Tests:*
Tested on docker.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
@@ -167,6 +167,7 @@ public class PerformanceAnalyzerApp {
     // what to do when the thread has stopped due to an unexpected exception.
     LOG.error("Thread: {} ran into an uncaught exception: {}", exception.getPaThreadName(),
         exception.getInnerThrowable());
+    exception.getInnerThrowable().printStackTrace();
     StatsCollector.instance().logException(exception.getExceptionCode());
   }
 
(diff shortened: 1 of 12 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
```diff
@@ -167,6 +167,7 @@ private static void handle(PAThreadException exception) {
     // what to do when the thread has stopped due to an unexpected exception.
     LOG.error("Thread: {} ran into an uncaught exception: {}", exception.getPaThreadName(),
         exception.getInnerThrowable());
+    exception.getInnerThrowable().printStackTrace();
```
yojs, line 170: Should we do this:
LOG.error("Thread: {} ran into an uncaught exception: {}", exception.getPaThreadName(),
        exception.getInnerThrowable(), exception);

to capture the stack trace in the logs ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/DummyGraph.java
```diff
@@ -125,9 +136,9 @@ public void construct() {
     hotNodeClusterRca.addTag(TAG_LOCUS, LOCUS_MASTER_NODE);
     hotNodeClusterRca.addAllUpstreams(Collections.singletonList(hotJVMNodeRca));
 
-    // constructResourceHeatMapGraph();
+    constructResourceHeatMapGraph();
```
yojs, line 139: If we are doing this, we might want to add it to muted RCA list. We don't want this enabled by default, right ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
ktkrg: I'll incorporate this.

ktkrg:
Closing this for a new one.

PR #184, opened 2020-05-05
Moving the configs to a conf file for HotShardRca and HotShardClusterRca

khushbr:
*Description of changes:* 
* Moving the threshold configs to conf files for HotShardRca and HotShardClusterRca
 * Updating the HotShardClusterRca to use median instead of average.
 * Adding metric for HotShardRca failure

### DIFF
--- a/pa_config/rca.conf
+++ b/pa_config/rca.conf
@@ -48,6 +48,12 @@
     "high-heap-usage-young-gen-rca": {
       "promotion-rate-mb-per-second" : 500,
       "young-gen-gc-time-ms-per-second" : 400
+    },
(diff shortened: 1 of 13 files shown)


pa_config/rca.conf
```diff
@@ -48,6 +48,12 @@
     "high-heap-usage-young-gen-rca": {
       "promotion-rate-mb-per-second" : 500,
       "young-gen-gc-time-ms-per-second" : 400
+    },
+    //hot shard rca
```
yojs, line 52: we can remove this comment.Its just the name of the json object

PR #186, opened 2020-05-06
Ut fix

khushbr:
*Description of changes:* 
* Adding `readAndUpdateMutesRcasDuringStart()` invoked during the controller `start()`. This method is invoked after graph initialization and before the RCASchedulerTask is initialized. This allows for a clean way to update the mutedRCA list within the Analysis graph. 
* Fixing the readAndUpdateMutedRcas() UT.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
@@ -55,17 +55,22 @@ import com.amazon.opendistro.elasticsearch.performanceanalyzer.threads.ThreadPro
 import com.sun.net.httpserver.HttpServer;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
+import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.sql.SQLException;
+import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Scanner;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -275,6 +275,17 @@ private void readRcaEnabledFromConf() {
         });
   }
 
+  private void readAndUpdateMutesRcasDuringStart() {
+    try {
+      Set<String> rcasForMute = new HashSet<>(rcaConf.getMutedRcaList());
+      rcasForMute.forEach(mutedRca -> Stats.getInstance().addToMutedGraphNodes(mutedRca));
```
yojs, line 281: I don't see that we are validating that there are Java classes with these names ?

yojs: Let's add a test that exercises this path. I mean when during the start of RcaController

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -150,7 +150,7 @@ private void start() {
       List<ConnectedComponent> connectedComponents = RcaUtil.getAnalysisGraphComponents(rcaConf);
 
       // Mute the rca nodes after the graph creation and before the scheduler start
-      readAndUpdateMutesRcas();
+      readAndUpdateMutesRcasDuringStart();
```
yojs, line 153: There is an edge case where both `readAndUpdateMutesRcas` and `readAndUpdateMutesRcasDuringStart` can try to update the muted Rca list back to back. So we would try to read in `rca.conf` twice. 

This will happen when rca was turned on and then off. Then between the two muted rcas update checks interval, some does both the things - adds a new muted rca and then turns on the rca. 

`readAndUpdateMutesRcasDuringStart` should ever be read at the start of the process itself. That's the time when RCA graph is not constructed and we cannot validate the new new muted RCAs. For any other updated to the muted list, the periodic rca.conf update checker will take care of it.

khushbr: Agreed, Have also added a comment explaining this.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
khushbr: > I don't see that we are validating that there are Java classes with these names ?

Done. Added check based on class name.

khushbr: > Let's add a test that exercises this path. I mean when during the start of RcaController

It is difficult to simulate this within a UT, since it relies on `lastModifiedTimeInMillisInMemory`, initialized at boot-up and `start()` being invoked. Will add an integration test to cover this.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -268,13 +268,35 @@ private void readRcaEnabledFromConf() {
...
+       * was turned off and then on.
+       *
+       * <p> `readAndUpdateMutesRcasDuringStart()` should only be read at the start of the process, when
+       * RCA graph is not constructed and we cannot validate the new new muted RCAs. For any other update
+       * to the muted list, the periodic rca.conf update checker will take care of it.
+       *
+       */
+      if (lastModifiedTimeInMillisInMemory == 0) {
+        Set<String> rcasForMute = new HashSet<>(rcaConf.getMutedRcaList());
+        rcasForMute.forEach(mutedRca -> Stats.getInstance().addToMutedGraphNodes(mutedRca));
```
yojs, line 291: Will it make sense to rewrite `updateMutedGraphNodes` as do two things:
- clear()
- addAll(nodeNames)

And just call that here instead of iterating through `rcasForMute` ?

khushbr: Yes, we can use the `updateMutedGraphNodes()` as is. Updating the code for the same.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -288,7 +288,7 @@ private void readAndUpdateMutesRcasDuringStart() {
        */
       if (lastModifiedTimeInMillisInMemory == 0) {
         Set<String> rcasForMute = new HashSet<>(rcaConf.getMutedRcaList());
-        rcasForMute.forEach(mutedRca -> Stats.getInstance().addToMutedGraphNodes(mutedRca));
+        Stats.getInstance().updateMutedGraphNodes(rcasForMute);
```
yojs, line 291: `updateMutedGraphNodes` retains the intersection of the elements in the original and what's provided if provided element is `nonEmpty` or else it adds them all. 

Because the muted rcas in the rca.conf is the final list, we should just clear the old elements and the new ones right ?

khushbr: The `mutedGraphNodes` is initialized as an empty set as part of the RCA Controller `start()`. 

Since `readAndUpdateMutesRcasDuringStart()`  is only invoked during the start, the `mutedGraphNodes` will be an empty set and the  `updateMutedGraphNodes()` on finding `mutedGraphNodes` as empty will add all the elements in rca.conf

yojs: `Stats` class is static. So, if RCA was disabled, muted rcas changed in the `rca.conf` and later enabled again, then based on this code, `updateMutedGraphNodes` will call `retainAll` that will be lead to an *intersection of the nodes* to be muted.

khushbr: Agreed, this condition was missed. Since rca.conf is our source of truth, now `updateMutedGraphNodes()` is clearing the older values in `mutedGraphNodes` and adding all from rca.conf.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -268,13 +268,35 @@ private void readRcaEnabledFromConf() {
...
+       * RCA graph is not constructed and we cannot validate the new new muted RCAs. For any other update
+       * to the muted list, the periodic rca.conf update checker will take care of it.
+       *
+       */
+      if (lastModifiedTimeInMillisInMemory == 0) {
+        Set<String> rcasForMute = new HashSet<>(rcaConf.getMutedRcaList());
+        Stats.getInstance().updateMutedGraphNodes(rcasForMute);
+        LOG.info("Updated the muted RCA Graph to : {}", rcaConf.getMutedRcaList());
+      }
+    } catch (Exception e) {
```
rguo-aws, line 294: Why do we need to handle exception here ? This make it quite confusing as we are not throwing any exceptions in between I assume ?

khushbr: We are only logging the exception with metric here and not throwing.

PR #192, opened 2020-05-14
Change the static summary table mapping to support 1 to n mapping

rguo-aws:
*Issue #, if available:*

*Description of changes:*
Change the static summary table mapping to support 1 to n mapping. Now the summary can have different types of summary object in its nested summary list. 
i.e.
We can attach both HotResourceSummary and HotShardSummary to HotNodeSummary. 

*Tests:*
tested on docker


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
```diff
@@ -218,42 +215,39 @@ public synchronized RcaResponse readRca(String rca) {
...
+    for (Class<? extends GenericSummary> clazz : clazzList) {
+      Field<Integer> foreignKeyField = DSL.field(
+          SQLiteQueryUtils.getPrimaryKeyColumnName(upperLevelTable), Integer.class);
+      SelectJoinStep<Record> rcaQuery = SQLiteQueryUtils
+          .buildSummaryQuery(create, clazz.getSimpleName(), upperLevelPrimaryKey, foreignKeyField);
+      try {
+        Result<Record> recordList = rcaQuery.fetch();
+        for (Record record : recordList) {
+          Method method = clazz.getMethod("buildSummary", Record.class);
+          GenericSummary summary = (GenericSummary) method.invoke(null, record);
```
khushbr, line 233: `(GenericSummary) method.invoke(null, record);`, is the null required here?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
```diff
@@ -218,42 +215,39 @@ public synchronized RcaResponse readRca(String rca) {
...
-      // it is totally fine if we fail to read some certain tables as some types of summaries might be missing
-      LOG.warn("Fail to read Summary table : {}, query = {},  exceptions : {}", currLevelTable, rcaQuery.toString(), de.getStackTrace());
+      catch (DataAccessException de) {
+        // it is totally fine if we fail to read some certain tables as some types of summaries might be missing
+        LOG.warn("Fail to read Summary table : {}, query = {},  exceptions : {}",
+            clazz.getSimpleName(), rcaQuery.toString(), de.getStackTrace());
+      } catch (Exception e) {
+        // we might got a reflection issue if running into this. Check the NestedTableMap and make sure
+        // the summary class has been added.
+        LOG.error("Fail to use reflection to build GenericSummary, trace = {}", e.getStackTrace());
```
khushbr, line 249: Consider adding a metric here.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
rguo-aws: yes. because buildSummary is a static method

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
rguo-aws: I don't think we need to add a metric here. we will only run into this if there is a bug in this code which doesn't needs to be shown up in the metric.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/SQLiteQueryUtils.java
```diff
@@ -48,24 +48,36 @@
...
+  static {
+    Map<Class<? extends GenericSummary>, List<Class<? extends GenericSummary>>> tableMap = new HashMap<>();
+    tableMap.put(RcaResponse.class, Collections.unmodifiableList(Collections.singletonList(
+        HotClusterSummary.class)));
+    tableMap.put(HotClusterSummary.class, Collections.unmodifiableList(Collections.singletonList(
+        HotNodeSummary.class)));
+    tableMap.put(HotNodeSummary.class, Collections.unmodifiableList(Collections.singletonList(
+        HotResourceSummary.class)));
+    tableMap.put(HotResourceSummary.class, Collections.unmodifiableList(Collections.singletonList(
+        TopConsumerSummary.class)));
```
yojs, line 72: I think the nesting should be internal to each top level summary rather than in the util class.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/SQLiteQueryUtils.java
```diff
@@ -48,24 +48,36 @@
...
+
+    //temperature profiling mapping
+    tableMap.put(ClusterTemperatureSummary.class, Collections.unmodifiableList(Collections.singletonList(
+        ClusterDimensionalSummary.class)));
     nestedTableMap = Collections.unmodifiableMap(tableMap);
   }
 
+
```
yojs, line 80: remove ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/SQLiteQueryUtils.java
rguo-aws: thanks for pointing this out. yeah this is indeed a much cleaner implementation. I have added the list to each summary class

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/SQLiteQueryUtils.java
rguo-aws: removed

PR #205, opened 2020-05-19
Prepare RCA to work with the Integration Test framework

sidheart:
*Issue #, if available:* 22

*Description of changes:* Modified the enablePaAndRca task to properly execute in stages. PA is enabled, then RCA is enabled. Previously this task would not always properly enable the two components.

Also modified gradle task logic to use a specified path to the docker-compose program if an environment variable is set. This allows the ES cluster to spin up on local machines as well as on the Github Actions CI runner.

*Tests:* N/A

*Code coverage percentage for this patch:* N/A

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -508,6 +508,28 @@ task unzipPerfTop(type: Copy) {
     enabled = !file(perfTopBin).exists()
 }
 
+task enablePa(type: Exec) {
+    dependsOn(waitForES)
+    doFirst {
+        sleep(5000)
+    }
+    commandLine('curl', 'localhost:9200/_opendistro/_performanceanalyzer/cluster/config', '-H',
+            'Content-Type: application/json', '-d', '{"enabled": true}')
+    doLast {
+        sleep(5000)
+    }
+}
+
+task enablePaAndRca(type: Exec) {
+    dependsOn(enablePa)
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -508,12 +523,25 @@ task unzipPerfTop(type: Copy) {
...
-task enablePaAndRca(type: Exec) {
+task enablePa(type: Exec) {
     dependsOn(waitForES)
+    doFirst {
+        sleep(5000)
+    }
     commandLine('curl', 'localhost:9200/_opendistro/_performanceanalyzer/cluster/config', '-H',
             'Content-Type: application/json', '-d', '{"enabled": true}')
+    doLast {
+        sleep(5000)
```
yojs, line 534: Should We add a comment explaining the rationale for the sleep ?

build.gradle
```diff
@@ -508,12 +523,25 @@ task unzipPerfTop(type: Copy) {
...
+    }
+}
+
+task enablePaAndRca(type: Exec) {
```
yojs, line 538: Should we call it `enableRca` ?

build.gradle
sidheart: Added

build.gradle
sidheart: Done

PR #221, opened 2020-06-03
collect queue latency metric from writer

rguo-aws:
*Issue #, if available:*

*Description of changes:*
collect queue latency metric from writer

*Tests:*
tested on docker


aditjind:
There is a circular dependency in the PA and RCA build jobs on GH. RCA build job on GH builds the PA as well but that build fails due to the metric JSON Key not present in the PA. Merge both of them at the same time and then the retry the failed job. Should pass on second attempt.

Issue #227, opened 2020-06-09
Restrict default access to service to localhost 


sruti1312:
Modify the default service access to only whitelist local address. 
Presently, it whitelists all remote addresses.

PR #228, opened 2020-06-09
Restrict default access to localhost

sruti1312:
*Issue #, if available:*
 Restrict default access to service to localhost #227 

*Description of changes:*
Restrict the default access to the service to localhost

*Tests:*
Unit test and Manually tested locally

### DIFF
--- a/pa_config/performance-analyzer.properties
+++ b/pa_config/performance-analyzer.properties
@@ -27,7 +27,7 @@ certificate-file-path = specify_path
 
 private-key-file-path = specify_path
 
-# WebService bind host; default to all interfaces
+# WebService bind host; default only to local interface
 #webservice-bind-host =
 
 # Plugin Stats Metadata file name, expected to be in the same location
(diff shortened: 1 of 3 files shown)


pa_config/performance-analyzer.properties
```diff
@@ -27,7 +27,7 @@ certificate-file-path = specify_path
 
 private-key-file-path = specify_path
 
-# WebService bind host; default to all interfaces
```
khushbr, line 30: Please also check if we need to update any public documentation for this.

dparker18:
Doesn't this change now force binding to localhost with no option to change it? Is there a way around that?

sruti1312:
> Doesn't this change now force binding to localhost with no option to change it? Is there a way around that?

Yes it enforces binding to localhost by default.  
You can follow https://opendistro.github.io/for-elasticsearch-docs/docs/install/plugins/#performance-analyzer to modify binding. 
With version 1.9.0 we have an issue around this, please refer to https://github.com/opendistro-for-elasticsearch/performance-analyzer/issues/154 on how to fix the issue. 
It will be fixed in 1.10 release.

dparker18:
Thank you I'll give that a try!

PR #252, opened 2020-06-18
Add NodeConfigCollector to collect node configs(threadpool capacity etc.) from ES

rguo-aws:
*Issue #, if available:*
#293 

*Description of changes:*
Add a base class EsConfigNode as node level config collector
Create derived class NodeConfigCollector to collect threadpool config settings from each node and wrap it as flowunit to send to cluster RCA node.
 
Create unit test for this collector

*Tests:*


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/remediation/NodeConfigurationRca.java
```diff
@@ -0,0 +1,93 @@
...
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.persist.SQLParsingUtil;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.HotNodeSummary;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.scheduler.FlowUnitOperationArgWrapper;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.reader.ClusterDetailsEventProcessor;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class NodeConfigurationRca extends Rca<ResourceFlowUnit<HotNodeSummary>> {
```
yojs, line 25: Please add a javadoc for this class

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/remediation/NodeConfigurationRca.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+    double searchQueueCapacity = SQLParsingUtil.readDataFromSqlResult(flowUnit.getData(),
+        THREAD_POOL_TYPE.getField(), ThreadPoolType.SEARCH.toString(), MetricsDB.MAX);
+    if (!Double.isNaN(searchQueueCapacity)) {
+      this.searchQueueCapacity = (int) searchQueueCapacity;
+    }
+  }
+
+  @Override
+  public ResourceFlowUnit<HotNodeSummary> operate() {
```
yojs, line 57: Let's add a javadoc stating how we are operating on the FlowUnits

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/remediation/NodeConfigurationRca.java
```diff
@@ -0,0 +1,93 @@
...
+public class NodeConfigurationRca extends Rca<ResourceFlowUnit<HotNodeSummary>> {
+
+  private static final Logger LOG = LogManager.getLogger(NodeConfigurationRca.class);
+  private final Metric threadPool_queueCapacity;
+  private final int rcaPeriod;
+  private int counter;
+  private int writeQueueCapacity;
+  private int searchQueueCapacity;
+
+  public <M extends Metric> NodeConfigurationRca(int rcaPeriod, M threadPool_queueCapacity) {
```
yojs, line 34: Can you briefly describe the rationale of using an RCA node for this ? Do you think it would be cleaner if we add a new node type called `ESConfigNode` ?

vigyasharma: @yojs are you suggesting that this should be different kind a `NonLeafNode` and not extend from RCA. With a FlowUnit that is more tuned to node configurations?

I think that is cleaner in general. Will need changes to PersistorBase logic which needs to be refactored anyway to work with non flow units that are not from RCAs. It can be picked separately.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/remediation/NodeConfigurationRca.java
rguo-aws: javadoc added

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/remediation/NodeConfigurationRca.java
rguo-aws: added

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/remediation/NodeConfigurationRca.java
rguo-aws: Created a new type of node EsConfigNode and extend the class as PerformanceControllerConfigCollector

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/EsConfigNode.java
```diff
@@ -0,0 +1,78 @@
...
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.scheduler.FlowUnitOperationArgWrapper;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * this is a base class for node(vertex) in RCA graph that reads configuration settings from ES.
+ */
+public abstract class EsConfigNode extends NonLeafNode<ResourceFlowUnit<HotNodeSummary>> {
```
vigyasharma, line 19: I was actually thinking of having a custom flow unit for this type of node. It doesn't really fit into the nested `HotNodeSummary` and `HotResourceSummary` lists - the terms healthy/unhealthy do not apply to a given queue capacity conf. The new flow unit could be simply have resource and values for those resources (for each node).

rguo-aws: Agree. that is actually what I am planing to do. but that requires refactoring in base cluster RCA. (i.e. need to create a base node level summary and extend it to something like EsConfigFlowunit). At this moment, we still have to add this to the HotNodeSummary so that it can be parsed by the cluster level RCA.  I will create an issue and refactor it altogether.

rguo-aws: Add NodeConfigFlowUnit as a new type of Flowunit to carry node config settings across RCA graph

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/flow_units/NodeConfigFlowUnit.java
```diff
@@ -0,0 +1,116 @@
...
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.contexts.ResourceContext;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.HotNodeSummary;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.HotResourceSummary;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.NodeKey;
+import java.util.HashMap;
+
+/**
+ * a flowunit type to carry ES node configurations (queue/cache capacities, etc.)
+ */
+public class NodeConfigFlowUnit extends ResourceFlowUnit<HotNodeSummary> {
```
yojs, line 32: Shouldn't the NodeConfig use its own summary such as `NodeConfigSummary` ?

rguo-aws: the problem of creating its own summary is this summary needs to be sent over to gRPC so we have to create a separate protobuf message for that. It would be better to keep the protobuf message as is to avoid duplicated code and create a new flowunit type as a wrapper to read/write config settings

PR #254, opened 2020-06-23
Add required mutual auth to gRPC Server/Client

sidheart:
*Issue #, if available:* 253

*Description of changes:*
Previously we had 1 sided TLS on the server side. Data between the
client and server was send over an encrypted channel, but any client
could make requests to the server.

This commit changes the behavior so that only clients with the matching
certificates can make requests to the server when TLS is enabled. This
commit does NOT add support for installing a trust manager. That must be
added in the future.

*Tests:*
- GRPCTest verifies that when https is enabled, an authenticated client is capable of making
requests, but an unauthenticated client is rejected.


src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCTest.java
```diff
@@ -0,0 +1,144 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.net;
```
yojs, line 1: This is awesome. Thanks for adding UTs around it

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/NetServer.java
```diff
@@ -234,5 +246,14 @@ public void stop() {
...
     subscribeHandler = null;
+
+    if (server != null) {
+      server.shutdown();
+      try {
+        server.awaitTermination(1, TimeUnit.MINUTES);
+      } catch (InterruptedException e) {
+        server.shutdownNow();
+      }
+    }
```
ktkrg, line 257: Are you sure you tested this change? We don't want to stop the server, it'll kill perftop and the ability to query metrics as well. If you want to change the behavior, I'd suggest we create a new issue and address this the right way in a new PR.

For the short term, you can consider renaming the method to removeRcaHandlers() or something like that.

gradle.properties
```diff
@@ -13,4 +13,5 @@
 #  permissions and limitations under the License.
 #
 
-localPaDir=../performance-analyzer
\ No newline at end of file
+localPaDir=../performance-analyzer
+org.gradle.jvmargs=-Xmx4096m -XX:MaxPermSize=256m
```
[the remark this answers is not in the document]
sidheart: Removed

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
```diff
@@ -145,12 +149,14 @@ private ManagedChannel buildInsecureChannel(final String remoteHost) {
...
+      File certFile = CertificateUtils.getCertificateFile();
+      File pkeyFile = CertificateUtils.getPrivateKeyFile();
       return NettyChannelBuilder.forAddress(remoteHost, Util.RPC_PORT)
                                 .sslContext(
                                     GrpcSslContexts.forClient()
-                                                   .trustManager(
-                                                       InsecureTrustManagerFactory.INSTANCE)
-                                                   .build())
+                                            .trustManager(InsecureTrustManagerFactory.INSTANCE)
+                                            .keyManager(certFile, pkeyFile)
```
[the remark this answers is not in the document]
sidheart: Done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/NetServer.java
sidheart: This is inside the stop method. I'm surprised that the stop method doesn't actually stop the server. Just in case we're relying on stop not to stop the server, I've removed the logic from stop() and added a shutdown() method which actually shuts everything down.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
```diff
@@ -140,26 +171,26 @@ private ManagedChannel buildChannelForHost(final String remoteHost) {
...
+               .sslContext(sslContextBuilder.build())
+               .build();
+     } catch (SSLException e) {
+       LOG.error("Unable to build an SSL gRPC client. Exception: {}", e.getMessage());
+       e.printStackTrace();
+
+       // Wrap the SSL Exception in a generic RTE and re-throw.
+       throw new RuntimeException(e);
+     }
+   }
```
ktkrg, line 193: nit: whitespace off by one.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
```diff
@@ -177,7 +208,15 @@ private void removeAllStubs() {
   private void terminateAllConnections() {
     for (Map.Entry<String, AtomicReference<ManagedChannel>> entry : perHostChannelMap.entrySet()) {
       LOG.debug("shutting down connection to host: {}", entry.getKey());
-      entry.getValue().get().shutdownNow();
+      ManagedChannel channel = entry.getValue().get();
+      channel.shutdownNow();
+      try {
+        channel.awaitTermination(1, TimeUnit.MINUTES);
+      } catch (InterruptedException e) {
```
ktkrg, line 215: You need to propagate the interrupted status when catching an InterruptedException by calling Thread.currentThread().interrupt() after channel.shutdownNow();

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/NetServer.java
```diff
@@ -235,4 +254,17 @@ public void stop() {
...
+
+  public void shutdown() {
+    stop();
+    // Actually stop the server
+    if (server != null) {
+      server.shutdown();
+      try {
+        server.awaitTermination(1, TimeUnit.MINUTES);
+      } catch (InterruptedException e) {
+        server.shutdownNow();
```
ktkrg, line 266: Same comment as above. You need to propagate the interrupted status by calling Thread.currentThread().interrupt();

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
sidheart: Fixed

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
sidheart: Done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/NetServer.java
sidheart: Done

PR #260, opened 2020-06-29
Decision Maker Skeleton

vigyasharma:
**Issue #, if available:** -N/A-

**Description of changes:** This PR defines the basic skeleton for the decision maker framework - a set of nodes that analyze RCA node observation data and create remediation action suggestions. Contains interfaces and base classes, along with simple pass-through implementations and analysis graph connections.

#### Components created:
1. `Decider`: Analysis data from multiple RCA nodes and creates a `Decision`, which is a set of candidate actions.
2. `Decision`: A set of remediation actions.
3. `Action`: Encapsulates a remediation operation along with the nodes it should be run against, and other execution parameters.
4. `Collator`: Subscribes to all deciders. Collates decisions from different deciders into a final set of decisions that are aligned in their impact on key resources. Prefers stability over performance.
5. `Publisher`: Applies appropriate damping functions on actions suggestions and executes them.

**Tests:** Added unit tests. For skeleton (pass-through) classes, tests will be added when they are properly implemented.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
@@ -1,14 +1,19 @@
 package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions;
 
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.NodeKey;
+
 import java.util.List;
 import java.util.Map;
 
 public interface Action {
 
     /**
-     * Constant to indicate that an action impacts all nodes in the cluster
+     * Returns true if the configured action is actionable, false otherwise.
+     *
+     * Examples of non-actionable actions are resource configurations where
+     * limits have been reached.
      */
-    final String ALL_NODES = "all_nodes";
+    boolean isActionable();
 
     /**
      * Time to wait since last recommendation, before suggesting this action again
@@ -16,25 +21,20 @@ public interface Action {
     int coolOffPeriodInSeconds();
 
     /**
-     * Returns a list of Elasticsearch NodeIds impacted by this action.
-     * {@link ALL_NODES} is used for cluster wide actions impacting all nodes.
-     */
-    List<String> impactedNodes();
-
-    /**
-     * Returns a map of Elasticsearch nodeId to ImpactVector of this action on that node
-     */
-    Map<String, ImpactVector> impact();
-
-    /**
-     * Returns Action Name
-     */
-    String getName();
-
-    /**
      * Called when the action is invoked.
      *
      * Specific implementation may include executing the action, or invoking downstream APIs
      */
     void execute();
+
+    /**
+     * Returns a list of Elasticsearch nodes impacted by this action.
+     */
+    List<NodeKey> impactedNodes();
+
+    /**
+     * Returns a map of Elasticsearch nodes to ImpactVector of this action on that node
+     */
+    Map<NodeKey, ImpactVector> impact();
+
 }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
@@ -31,6 +31,9 @@ public interface Action {
   /** Returns a map of Elasticsearch nodes to ImpactVector of this action on that node */
   Map<NodeKey, ImpactVector> impact();
 
-  /** Return action name */
+  /** Returns action name */
   String name();
+
+  /**  Returns a summary for the configured action */
+  String summary();
 }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
@@ -1,6 +1,14 @@
 package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions;
 
+import java.util.List;
+import java.util.Map;
+
 public interface Action {
+
+    /**
+     * Constant to indicate that an action impacts all nodes in the cluster
+     */
+    final String ALL_NODES = "all_nodes";
 
     /**
      * Time to wait since last recommendation, before suggesting this action again
@@ -8,9 +16,15 @@ public interface Action {
     int coolOffPeriodInSeconds();
 
     /**
-     * Returns an ImpactVector indicating the impact of this action on key resources.
+     * Returns a list of Elasticsearch NodeIds impacted by this action.
+     * {@link ALL_NODES} is used for cluster wide actions impacting all nodes.
      */
-    ImpactVector impact();
+    List<String> impactedNodes();
+
+    /**
+     * Returns a map of Elasticsearch nodeId to ImpactVector of this action on that node
+     */
+    Map<String, ImpactVector> impact();
 
     /**
      * Returns Action Name
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
@@ -1,0 +1,26 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions;
+
+public interface Action {
+
+    /**
+     * Time to wait since last recommendation, before suggesting this action again
+     */
+    int coolOffPeriodInSeconds();
+
+    /**
+     * Returns an ImpactVector indicating the impact of this action on key resources.
+     */
+    ImpactVector impact();
+
+    /**
+     * Returns Action Name
+     */
+    String getName();
+
+    /**
+     * Called when the action is invoked.
+     *
+     * Specific implementation may include executing the action, or invoking downstream APIs
+     */
+    void execute();
+}
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
@@ -7,39 +7,30 @@ import java.util.Map;
 
 public interface Action {
 
-    /**
-     * Returns true if the configured action is actionable, false otherwise.
-     *
-     * Examples of non-actionable actions are resource configurations where
-     * limits have been reached.
-     */
-    boolean isActionable();
+  /**
+   * Returns true if the configured action is actionable, false otherwise.
+   *
+   * <p>Examples of non-actionable actions are resource configurations where limits have been
+   * reached.
+   */
+  boolean isActionable();
 
-    /**
-     * Time to wait since last recommendation, before suggesting this action again
-     */
-    int coolOffPeriodInSeconds();
+  /** Time to wait since last recommendation, before suggesting this action again */
+  int coolOffPeriodInSeconds();
 
-    /**
-     * Called when the action is invoked.
-     *
-     * Specific implementation may include executing the action, or invoking downstream APIs
-     */
-    void execute();
+  /**
+   * Called when the action is invoked.
+   *
+   * <p>Specific implementation may include executing the action, or invoking downstream APIs
+   */
+  void execute();
 
-    /**
-     * Returns a list of Elasticsearch nodes impacted by this action.
-     */
-    List<NodeKey> impactedNodes();
+  /** Returns a list of Elasticsearch nodes impacted by this action. */
+  List<NodeKey> impactedNodes();
 
-    /**
-     * Returns a map of Elasticsearch nodes to ImpactVector of this action on that node
-     */
-    Map<NodeKey, ImpactVector> impact();
+  /** Returns a map of Elasticsearch nodes to ImpactVector of this action on that node */
+  Map<NodeKey, ImpactVector> impact();
 
-    /**
-     * Return action name
-     */
-    String name();
-
+  /** Return action name */
+  String name();
 }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
 package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions;
 
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.NodeKey;
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
@@ -49,6 +49,6 @@ public interface Action {
   /** Returns action name */
   String name();
 
-  /**  Returns a summary for the configured action */
+  /** Returns a summary for the configured action */
   String summary();
 }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
@@ -37,4 +37,9 @@ public interface Action {
      */
     Map<NodeKey, ImpactVector> impact();
 
+    /**
+     * Return action name
+     */
+    String name();
+
 }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ImpactVector.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ImpactVector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ImpactVector.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ImpactVector.java
@@ -1,0 +1,51 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ImpactVector {
+
+    public enum Resource {
+        HEAP,
+        CPU,
+        RAM,
+        DISK,
+        NETWORK
+    }
+
+    public enum Impact {
+        NO_IMPACT,
+        INCREASES_PRESSURE,
+        DECREASES_PRESSURE
+    }
+
+    private Map<Resource, Impact> impactMap = new HashMap<>();
+
+    public ImpactVector() {
+        for (Resource r: Resource.values()) {
+            impactMap.put(r, Impact.NO_IMPACT);
+        }
+    }
+
+    public Map<Resource, Impact> getImpact() {
+        return impactMap;
+    }
+
+    public void increasesPressure(Resource... resources) {
+        for (Resource resource: resources) {
+            impactMap.put(resource, Impact.INCREASES_PRESSURE);
+        }
+    }
+
+    public void decreasesPressure(Resource... resources) {
+        for (Resource resource: resources) {
+            impactMap.put(resource, Impact.DECREASES_PRESSURE);
+        }
+    }
+
+    public void noImpact(Resource... resources) {
+        for (Resource resource: resources) {
+            impactMap.put(resource, Impact.NO_IMPACT);
+        }
+    }
+}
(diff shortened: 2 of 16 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ImpactVector.java
```diff
@@ -0,0 +1,51 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ImpactVector {
+
+    public enum Resource {
```
rguo-aws, line 8: should we rename this enum to something like "ResourceVector" ? we already have a Resource class defined in protobuf

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/QueueCapacity.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Map;
+
+import static com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.ImpactVector.Resource.CPU;
+import static com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.ImpactVector.Resource.HEAP;
+import static com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.ImpactVector.Resource.NETWORK;
+
+public class QueueCapacity implements Action {
+
+    public static final String NAME = "queue_capacity";
+    public static final int COOL_OFF_PERIOD = 300;
```
rguo-aws, line 18: rename this variable to "COOL_OFF_PERIOD_IN_SECOND" ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public QueueHealthDecider(long evalIntervalSeconds, int decisionFrequency, QueueRejectionClusterRca queueRejectionClusterRca) {
+        // TODO: Also consume NodeConfigurationRca
+        super(evalIntervalSeconds, decisionFrequency);
+        this.queueRejectionRca = queueRejectionClusterRca;
+        configureActionPriority();
+    }
+
+    @Override
+    public String name() {
```
rguo-aws, line 39: is there any specific reason to override this method ? The default implementation is to return getClass().getSimpleName() which will return "QueueHealthDecider" if called from this class.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
```diff
@@ -0,0 +1,64 @@
...
+    private Collator collator;
+    private boolean isMuted = false;
+
+    public Publisher(int evalIntervalSeconds, Collator collator) {
+        super(0, evalIntervalSeconds);
+        this.collator = collator;
+    }
+
+    @Override
+    public GenericFlowUnit operate() {
```
rguo-aws, line 24: maybe we can create a dedicated EmptyFlowUnit for the top level vertex in graph ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ImpactVector.java
vigyasharma: Renamed to `Dimension`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
vigyasharma: If we maintain a registry of all graph objects going forward, to allow a refactor as well as pluggable graph nodes, then the `public static NAME` field will serve as a key. For comparisons in that case, each node should return their `NAME` variable in `name()`. Hence the override

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
vigyasharma: I'm of split opinion here, but created one so as to explicitly identify the terminal node.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
```diff
@@ -0,0 +1,39 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions;
```
rguo-aws, line 1: let's add the copyright header on top of each newly created file

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/QueueCapacity.java
```diff
@@ -0,0 +1,135 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class QueueCapacity implements Action {
+
+  public static final String NAME = "queue_capacity";
+  public static final int COOL_OFF_PERIOD_IN_SECONDS = 300;
```
yojs, line 34: Do we want to make it configurable ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/QueueCapacity.java
```diff
@@ -0,0 +1,135 @@
...
+  @Override
+  public List<NodeKey> impactedNodes() {
+    return Collections.singletonList(esNode);
+  }
+
+  @Override
+  public Map<NodeKey, ImpactVector> impact() {
+    ImpactVector impactVector = new ImpactVector();
+    if (desiredCapacity > currentCapacity) {
+      impactVector.increasesPressure(HEAP, CPU, NETWORK);
```
yojs, line 78: Is it true that increasing the queue capacity will increase the CPU load ? You mean for CPU bound processing of queue items ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Collator.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public class Collator extends Decider {
+
+  public static final String NAME = "collator";
+
+  /* Deciders can choose to publish decisions at different frequencies based on the
+   * type of resources monitored and rca signals. The collator should however, not introduce any
+   * unnecessary delays. As soon as a decision is taken, it should be evaluated and published downstream.
+   */
+  private static final int collatorFrequency = 1;
```
yojs, line 43: Should we add the unit to the variable name ?

vigyasharma:
Did a merge from master instead of rebase, which has increased the number of files. The new changes are essentially around making it compatible with the new protobuf. 
Code changes are up to commit -- 38c22799d7089e4ab14f8951eca154a57713d234

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ImpactVector.java
```diff
@@ -0,0 +1,66 @@
...
+    HEAP,
+    CPU,
+    RAM,
+    DISK,
+    NETWORK
+  }
+
+  public enum Impact {
+    NO_IMPACT,
+    INCREASES_PRESSURE,
```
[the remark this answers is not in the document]
vigyasharma: I wanted to remove the ambiguity around whether it increases pressure (consumes more resource) or increases available resource (frees up the resource).

e.g. `HEAP, INCREASES` may imply it increases the overall available heap on the node.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/QueueCapacity.java
```diff
@@ -0,0 +1,135 @@
...
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.ThreadPoolEnum;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.NodeKey;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class QueueCapacity implements Action {
```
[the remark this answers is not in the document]
vigyasharma: My initial thoughts were that the prefix verb will be same across many actions. Most of them will be `Change*` or `Modify*` (except moveShard etc actions), which seemed redundant. But it does sound more complete, so we can change it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/QueueCapacity.java
vigyasharma: Since it increases throughput, I think it would increase the CPU load. 
CPU ends up doing more amortized work per unit time. Bursty requests that would've otherwise been rejected and retried, are now accepted by the larger queue and processed; making cpu less idle and increasing throughput.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/QueueCapacity.java
vigyasharma: This timeout is independent of its calling function, so we shouldn't need it configurable via callers. Did you mean reading and setting this value from rca.conf? I'm not sure if we'll want to edit this at runtime (maybe we can make that change when we add rca.conf integration to this class, or if we see a need later).

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -0,0 +1,135 @@
...
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class ModifyQueueCapacityAction implements Action {
+
+  public static final String NAME = "modify_queue_capacity";
+  public static final int COOL_OFF_PERIOD_IN_SECONDS = 300;
```
sendkb, line 33: it may be ok for the first cut, but we may have to make this param configurable for different actions

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -0,0 +1,135 @@
...
+  public void execute() {
+    // Making this a no-op for now
+    // TODO: Modify based on downstream agent API calls
+    assert true;
+  }
+
+  @Override
+  public String summary() {
+    if (!isActionable()) {
+      return "No action to take";
```
sendkb, line 94: nit: 'No Queue action to take'

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sendkb, line 2: nit: all license headers should have 2020 as the year

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -0,0 +1,135 @@
...
+  private Map<ResourceEnum, Integer> lowerBound = new HashMap<>();
+  private Map<ResourceEnum, Integer> upperBound = new HashMap<>();
+
+  public ModifyQueueCapacityAction(NodeKey esNode, ResourceEnum threadPool, int currentCapacity, boolean increase) {
+    setBounds();
+    int STEP_SIZE = 50;
+    this.esNode = esNode;
+    this.threadPool = threadPool;
+    this.currentCapacity = currentCapacity;
+    int desiredCapacity = increase ? currentCapacity + STEP_SIZE : currentCapacity - STEP_SIZE;
```
sendkb, line 49: we only consider increase or decrease of queue resources here. Is it safe to assume that an obj will not be created at all if there is no change in values ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -0,0 +1,135 @@
...
+    this.lowerBound.put(ResourceEnum.WRITE_THREADPOOL, 100);
+    this.upperBound.put(ResourceEnum.WRITE_THREADPOOL, 1000);
+
+    // Search thread pool
+    this.lowerBound.put(ResourceEnum.SEARCH_THREADPOOL, 1000);
+    this.upperBound.put(ResourceEnum.SEARCH_THREADPOOL, 3000);
+  }
+
+  private void setDesiredCapacity(int desiredCapacity) {
+    this.desiredCapacity = Math.min(desiredCapacity, upperBound.get(threadPool));
```
sendkb, line 120: minor: the two lines can be simplified into a single one. Change it if you think the readability is not affected

Math.min(Math.max(desiredCapacity, lowerBound.get(threadPool)), upperBound.get(threadPool))

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
vigyasharma: Its not clear what we mean by configurable here -

1. Configurable across different instances of `ModifyQueueCapacityAction` ? 
I think it would be the same for a particular type of action, we would want to wait `x` amount of time when modifying queues. Possibly different wait times for increase v/s decrease queue capacity, but that logic (waiting for queue drain out) can lie within `coolOffPeriodInSeconds()`. I'm not sure why different instances of `ModifyQueueCapacityAction` will need different cool off periods.

2. Configurable across different types of action, like modifyCacheBound/modifyQueueCapacity etc. -
Each such action class will have its own cool off period defined.

3. Configurable via `rca.conf` -
We can add this logic in `coolOffPeriodInSeconds` when needed, but I'm still skeptical if we want to add another knob in `rca.conf`. It gets hard to manually maintain all these configurations. Instead, the function might return a cool off based on some calculations, like cool off for moveShard may depend on shard size and recovery speed configured, rather than a manual setting per cluster.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
vigyasharma: Actions are created by deciders only when they need to be triggered, and they always do something; hence the assumption.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/QueueCapacity.java
yojs: Ya, I meant it reading from rca.conf. If you want to do an A/B test to come up with optimal value, keeping it configurable might help. Otherwise, we will have to deploy jars with the changes.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/QueueCapacity.java
yojs: Sorry about that, my point was because this is a PR for the framework, I am suggesting that what the decider predicts about resource usage might be taken as a hint and not as a definitive statement. 
But I guess, this can be a future improvement for the learning system.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
yojs: These are good points. We should add them to the class level javadoc

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -0,0 +1,134 @@
...
+  @Override
+  public void execute() {
+    // Making this a no-op for now
+    // TODO: Modify based on downstream agent API calls
+    assert true;
+  }
+
+  @Override
+  public String summary() {
+    if (!isActionable()) {
```
yojs, line 93: As stepsize is a positive integer, the creation of an action means that the desired size `!=` currentCapacity as here:
`desiredCapacity = increase ? currentCapacity + STEP_SIZE : currentCapacity - STEP_SIZE;`

What are the cases when it is not actionable ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -0,0 +1,134 @@
...
+      return String.format("No action to take for: [%s]", NAME);
+    }
+    return String.format("Update [%s] queue capacity from [%d] to [%d] on node [%s]",
```
yojs, line 96: `String.format` uses complicated pattern matching to create the final string, do you think we can use StringBuilder ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decision.java
```diff
@@ -0,0 +1,69 @@
...
+      actions.add(action);
+    }
+  }
+
+  public void addAllActions(List<Action> actions) {
+    this.actions.addAll(actions);
+  }
+
+  public List<Action> getActions() {
+    return actions;
```
yojs, line 46: Do we want to send an `unmodifiableList` list ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
```diff
@@ -0,0 +1,72 @@
...
+
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.deciders;
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.Action;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.core.NonLeafNode;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.scheduler.FlowUnitOperationArgWrapper;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class Publisher extends NonLeafNode<EmptyFlowUnit> {
```
yojs, line 24: a class level javadoc comment ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Override
+  public EmptyFlowUnit operate() {
+    // TODO: Pass through implementation, need to add dampening, cool-off, action flip-flop
+    // avoidance, state persistence etc.
+
+    Decision decision = collator.getFlowUnits().get(0);
+    for (Action action : decision.getActions()) {
+      LOG.info("Executing action: [{}]", action.name());
```
yojs, line 43: Do you think we can add a debug message and a metric here ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+// This is a sample decider implementation to finalize decision maker interfaces.
+// TODO: 1. Read action priorities from a configurable yml
+// TODO: 2. Read current queue capacity from NodeConfigurationRca (PR #252)
+
+public class QueueHealthDecider extends Decider {
+
+  public static final String NAME = "queue_health";
```
yojs, line 36: We already have a camelCase and mixed snake case in [namings](https://opendistro.github.io/for-elasticsearch-docs/docs/pa/reference/), now this will be pure snake-case. Do we want to stick to one of the previous two or maybe camelCase ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -0,0 +1,125 @@
...
+    if (counter < decisionFrequency) {
+      return decision;
+    }
+
+    counter = 0;
+    if (queueRejectionRca.getFlowUnits().isEmpty()) {
+      return decision;
+    }
+
+    HotClusterSummary clusterSummary = queueRejectionRca.getFlowUnits().get(0).getSummary();
```
yojs, line 67: `queueRejectionRca` can be empty in some cases, which means that the summary might be `null`. We might want to make an early return in that case ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -0,0 +1,125 @@
...
+  private Action getAction(String actionName, NodeKey esNode, ResourceEnum threadPool, int currCapacity, boolean increase) {
+    switch (actionName) {
+      case ModifyQueueCapacityAction.NAME:
+        return configureQueueCapacity(esNode, threadPool, currCapacity, increase);
+      default:
+        return null;
+    }
+  }
+
+  private ModifyQueueCapacityAction configureQueueCapacity(NodeKey esNode, ResourceEnum threadPool, int currentCapacity, boolean increase) {
```
yojs, line 110: Because this method can return `null`, do we want to make it explicit using the annotation `@nullable` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElasticSearchAnalysisGraph.java
```diff
@@ -150,8 +157,40 @@ public void construct() {
...
+    // Cluster level queue rejection RCA
+    QueueRejectionClusterRca queueRejectionClusterRca = new QueueRejectionClusterRca(RCA_PERIOD, queueRejectionNodeRca);
+    queueRejectionClusterRca.addTag(TAG_LOCUS, LOCUS_MASTER_NODE);
+    queueRejectionClusterRca.addAllUpstreams(Collections.singletonList(queueRejectionNodeRca));
+
+    // Queue Health Decider
+    QueueHealthDecider queueHealthDecider = new QueueHealthDecider(EVALUATION_INTERVAL_SECONDS, 12, queueRejectionClusterRca);
+    queueHealthDecider.addTag(TAG_LOCUS, LOCUS_MASTER_NODE);
+    queueHealthDecider.addAllUpstreams(Collections.singletonList(queueRejectionClusterRca));
+
```
yojs, line 181: can we extract out the code here in a method of its own as in `constructShardResourceUsageGraph` ?

yojs:
It might be good to add a readme in the doc folder with a brief description of each of the components of the decision maker - 
deciders, decisions, collators, publishers etc. with a note on what their locus of execution is - data/master node.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/QueueCapacity.java
vigyasharma: Let's discuss and track this separately. If needed, we can add this along with the other rca.conf integrations which are pending.

vigyasharma:
Thanks for the review @yojs. Merged in the changes to unblock some pending tasks. I'll address your suggestions in a separate PR.

PR #265, opened 2020-07-07
FieldData and Shard Request Cache RCA

khushbr:
*Issue #, if available:* -N/A-

*Description of changes:* This PR adds changes for Field Data Cache and Shard Request Cache RCA.
 
1. **Field Data Cache RCA** is to identify when the cache is unhealthy(thrashing) and otherwise, healthy. The dimension we are using for this analysis is cache eviction count, cache current weight(size) and cache max weight(size) configured.
Note : For Field Data Cache, Hit and Miss metrics aren't available.

2. **Shard Request Cache RCA** is to identify when the cache is unhealthy(thrashing) and otherwise, healthy. The dimension we are using for this analysis is cache eviction, hit count, cache current weight(size) and cache max weight(size) configured.

*Tests:* Unit Tests present.


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/FieldDataCacheRca.java
```diff
@@ -0,0 +1,204 @@
...
+        cacheEvictionCollector.collect(currTimestamp);
+        if (counter >= rcaPeriod) {
+            ResourceContext context;
+            HotNodeSummary nodeSummary;
+
+            ClusterDetailsEventProcessor.NodeDetails currentNode = ClusterDetailsEventProcessor.getCurrentNodeDetails();
+            double cacheSize = getTotalSizeInMB(fieldDataCacheSize);
+            double cacheMaxSize = getTotalSizeInMB(fieldDataCacheMaxSize);
+            exceedsSize = cacheMaxSize != 0 && cacheMaxSize != 0 && cacheSize > cacheMaxSize;
+            if (cacheEvictionCollector.isUnhealthy(currTimestamp) && exceedsSize) {
```
vigyasharma, line 114: Do we need the `exceedsSize` check for unhealthy? It's not clear to me why.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/FieldDataCacheRca.java
```diff
@@ -0,0 +1,204 @@
...
+            this.evictionTimePeriodThreshold = threshold;
+        }
+
+        public void collect(final long currTimestamp) {
+            for (MetricFlowUnit flowUnit : cacheEvictionMetrics.getFlowUnits()) {
+                if (flowUnit.isEmpty() || flowUnit.getData() == null) {
+                    continue;
+                }
+
+                double evictionCount = flowUnit.getData().stream().mapToDouble(
```
vigyasharma, line 170: `flowUnit.getData().stream()` -- Can each flowUnit here have multiple data points?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/ShardRequestCacheRca.java
```diff
@@ -0,0 +1,223 @@
...
+ * The dimension we are using for this analysis is cache eviction, hit count, cache current weight(size)
+ * and cache max weight(size) configured.
+ *
+ * <p>Cache eviction within Elasticsearch happens in following scenarios:
+ * <ol>
+ *   <li> Mutation to Cache (Entry Insertion/Promotion and Manual Invalidation)
+ *   <li> Explicit call to refresh()
+ * </ol>
+ *
+ * <p>The Cache Eviction requires that either the cache weight exceeds OR the entry TTL is expired.
```
vigyasharma, line 57: Same as above

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/ShardRequestCacheRca.java
```diff
@@ -0,0 +1,223 @@
...
+        cacheEvictionCollector.collect(currTimestamp);
+        cacheHitCollector.collect(currTimestamp);
+        if (counter >= rcaPeriod) {
+            ResourceContext context;
+            HotNodeSummary nodeSummary;
+
+            ClusterDetailsEventProcessor.NodeDetails currentNode = ClusterDetailsEventProcessor.getCurrentNodeDetails();
+            double cacheSize = getTotalSizeInMB(shardRequestCacheSize);
+            double cacheMaxSize = getTotalSizeInMB(shardRequestCacheMaxSize);
+            exceedsSize = cacheMaxSize != 0 && cacheMaxSize != 0 && cacheSize > cacheMaxSize;
```
vigyasharma, line 128: same as above

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/ShardRequestCacheRca.java
```diff
@@ -0,0 +1,223 @@
...
+
+            // if eviction and hit counts persists in last 5 minutes and cache size exceeds the max cache size configured,
+            // the cache is considered as unhealthy
+            if (cacheEvictionCollector.isMetricPresentForThresholdTime(currTimestamp)
+                    && cacheHitCollector.isMetricPresentForThresholdTime(currTimestamp)
+                    && exceedsSize) {
```
vigyasharma, line 134: It's not clear why `exceedsSize` is important for thrashing. 
Is this to remove false positives due to TTL based evictions? The doc string mentions that TTL based evictions are not performed because setting is not present and so we ignore it.

Even to remove the TTL eviction false +ves, cacheSize may not *exceed* maxCacheSize because -
1. ES accounting should mostly be accurate, which prevents cacheSize from exceeding cacheMaxSize. 
2. Evictions should have brought down the cacheSize.

Maybe we should check for cacheSize to be within a threshold of cacheMaxSize: assume evictions were due to TTL if `cacheMaxSize - cacheSize > Threshold`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/ShardRequestCacheRca.java
```diff
@@ -0,0 +1,223 @@
...
+            exceedsSize = Boolean.FALSE;
+            return new ResourceFlowUnit<>(currTimestamp, context, nodeSummary, !currentNode.getIsMasterNode());
+        }
+        else {
+            return new ResourceFlowUnit<>(currTimestamp);
+        }
+    }
+
+    @Override
+    public void generateFlowUnitListFromWire(FlowUnitOperationArgWrapper args) {
```
vigyasharma, line 153: same as above

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/ShardRequestCacheRca.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        public void collect(final long currTimestamp) {
+            for (MetricFlowUnit flowUnit : cacheMetrics.getFlowUnits()) {
+                if (flowUnit.isEmpty()) {
+                    continue;
+                }
+
+                Result<Record> records = flowUnit.getData();
+                double metricCount = records.stream().mapToDouble(
```
vigyasharma, line 189: same as above

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/FieldDataCacheRca.java
```diff
@@ -0,0 +1,204 @@
...
+
+        cacheEvictionCollector.collect(currTimestamp);
+        if (counter >= rcaPeriod) {
+            ResourceContext context;
+            HotNodeSummary nodeSummary;
+
+            ClusterDetailsEventProcessor.NodeDetails currentNode = ClusterDetailsEventProcessor.getCurrentNodeDetails();
+            double cacheSize = getTotalSizeInMB(fieldDataCacheSize);
+            double cacheMaxSize = getTotalSizeInMB(fieldDataCacheMaxSize);
+            exceedsSize = cacheMaxSize != 0 && cacheMaxSize != 0 && cacheSize > cacheMaxSize;
```
sruti1312, line 113: We are checking if cacheMaxSize is not zero twice.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/CacheUtil.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class CacheUtil {
+    private static final Logger LOG = LogManager.getLogger(CacheUtil.class);
+    private static final double CONVERT_BYTES_TO_MEGABYTES = Math.pow(1024, 3);
+
+    public static Double getTotalSizeInMB(final Metric sizeMetric) {
+        double sizeTotalInMB = 0;
+
+        // we expect the Metric to have single flow unit since it is consumed locally
+        MetricFlowUnit flowUnit = sizeMetric.getFlowUnits().get(0);
```
rguo-aws, line 33: should we check the size the flowunit list to avoid null pointer exception ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/CacheUtil.java
```diff
@@ -0,0 +1,48 @@
...
+        if (flowUnit.isEmpty() || flowUnit.getData() == null) {
+            return sizeTotalInMB;
+        }
+
+        for (Record record : flowUnit.getData()) {
+            double size = record.getValue(MetricsDB.MAX, Double.class);
```
rguo-aws, line 39: instead of iterating through the list of tuples in record, can we extend the cache metric class and leverage the SQL group by statement when reading from the db ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/FieldDataCacheRca.java
```diff
@@ -0,0 +1,204 @@
...
+            if (cacheEvictionCollector.isUnhealthy(currTimestamp) && exceedsSize) {
+                context = new ResourceContext(Resources.State.UNHEALTHY);
+                nodeSummary = new HotNodeSummary(currentNode.getId(), currentNode.getHostAddress());
+                nodeSummary.appendNestedSummary(cacheEvictionCollector.generateSummary(currTimestamp));
+            }
+            else {
+                context = new ResourceContext(Resources.State.HEALTHY);
+                nodeSummary = null;
```
rguo-aws, line 121: let's add node summary to healthy flowunit and send it as a heartbeat to cluster rca

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/FieldDataCacheRca.java
rguo-aws: agree, and I don't think we need exceedsSize here. And this value will immediately jump to non-zero when a scaling down action has been applied so it is not a good indicator for cache healthiness

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/FieldDataCacheRca.java
```diff
@@ -0,0 +1,204 @@
...
+    private final Metric fieldDataCacheSize;
+    private final Metric fieldDataCacheMaxSize;
+    private final int rcaPeriod;
+    private int counter;
+    private boolean exceedsSize;
+    protected Clock clock;
+    private final CacheEvictionCollector cacheEvictionCollector;
+
+    public <M extends Metric> FieldDataCacheRca(final int rcaPeriod, final M fieldDataCacheEvictions,
+                                                final M fieldDataCacheSize, final M fieldDataCacheMaxSize) {
```
rguo-aws, line 82: fieldDataCacheMaxSize will be collected as part node node config RCA. Do we need a separate metric here ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/ShardRequestCacheRca.java
```diff
@@ -0,0 +1,223 @@
...
+ * <p>The Cache Hit and Eviction metric presence implies cache is undergoing frequent load and eviction or undergoing
+ * scheduled cleanup for entries which had timed-out during execution.
+ *
+ * <p>This RCA reads 'shardRequestCacheEvictions',  'shardRequestCacheHits', 'shardRequestCacheSize' and
+ * 'shardRequestCacheMaxSize' from upstream metrics and maintains collectors which keeps track of the time window
+ * period(tp) where we repeatedly see evictions and hits for the last tp duration. This RCA is marked as unhealthy
+ * if tp we find tp is above the threshold(300 seconds) and cache size exceeds the max cache size configured.
+ *
+ */
+public class ShardRequestCacheRca extends Rca<ResourceFlowUnit<HotNodeSummary>> {
```
rguo-aws, line 76: can we combine ShardRequestCacheRca and FieldDataCacheRca into one RCA and use different collector to collect those caches individually to reduce overhead ? Is there any benefit of creating two separate RCAs ?

src/main/proto/inter_node_rpc_service.proto
```diff
@@ -69,6 +69,11 @@ enum ResourceEnum {
   // threadpool
   WRITE_THREADPOOL = 4 [(additional_fields).name = "write threadpool"];
   SEARCH_THREADPOOL = 5 [(additional_fields).name = "search threadpool"];
+
+  // Cache
+  FIELD_DATA_CACHE = 6 [(additional_fields).name = "field data cache"];
```
rguo-aws, line 74: let's reserve some field numbers for thread pool and many be we can start from 10 here ?

src/main/proto/inter_node_rpc_service.proto
```diff
@@ -84,6 +89,10 @@ enum MetricEnum {
 
   // threadpool
   QUEUE_REJECTION = 6 [(additional_fields).name = "queue rejection", (additional_fields).description = "rejection period in second"];
+
+  // cache
+  CACHE_EVICTION = 7 [(additional_fields).name = "cache eviction", (additional_fields).description = "cache eviction count"];
```
rguo-aws, line 94: same as above, let reserve some field number for queue latency metric etc.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/FieldDataCacheRca.java
vigyasharma: Got it, so we need to sum the stream because this is a shard level metric?

PR #270, opened 2020-07-13
Adding cache max size metrics 

khushbr:
*Issue #, if available:* N/A

*Description of changes:*  Adding code change for collecting cache max size metrics for Field Data and Shard Request Cache.
```
sqlite> select * from Cache_MaxSize;
Field Data Cache|26214400.0|26214400.0|26214400.0|26214400.0
Shard Request Cache|80181985.0|80181985.0|80181985.0|80181985.0
sqlite>
```

*Tests:*  Pending UTs, I will add them as part of next iteration once we have the collector name and other variables(Cache Table Name, Dimension and Value) finalized.


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metrics/AllMetrics.java
```diff
@@ -32,6 +32,7 @@
   // metric name (not complete, only metrics use the json format and contains
   // numeric values. Will add more when needed)
   public enum MetricName {
+    CACHE_CUSTOM,
```
vigyasharma, line 35: Minor: The naming is not very intuitive for me. Should this just be `CACHE METRICS`? Maybe I'm missing the context around how it is intended to be used.

khushbr: Adding the writer side PR for more context : https://github.com/opendistro-for-elasticsearch/performance-analyzer/pull/145

The idea is that cache max size is a node level static setting, unlike cache hit, miss, size and eviction (part of SHARD_STATS) which are per shard and thus, the two category cannot be clubbed together and  `CACHE METRICS` being the generic name cannot be used. 

I agree `CACHE_CUSTOM` isnt very explanatory and I am open to better naming suggestions here.

vigyasharma: Thanks for the pointers. How about `CACHE_CONFIG`, since this has cache configuration (max_size etc.) ?

rguo-aws: can we simple call it CACHE ? There is also a THREAD_POOL metric down below. I think we can simply call it CACHE. 

And why can't we add this directly into SHARD_STATS metric ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/metrics/Cache_Max_Size.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.metrics;
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.metrics.AllMetrics;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.Metric;
+
+public class Cache_Max_Size extends Metric {
+
+    public static final String NAME = AllMetrics.CacheCustomValue.CACHE_MAX_SIZE.name();
```
rguo-aws, line 23: should we use .toString() here ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metrics/AllMetrics.java
```diff
@@ -129,6 +130,44 @@ public String toString() {
...
+      this.value = value;
+    }
+
+    @Override
+    public String toString() {
+      return value;
+    }
+
+    public static class Constants {
+      public static final String MAX_SIZE = "Cache_MaxSize";
```
rguo-aws, line 167: should we follow the  naming convention  and rename this to CACHE_MAX_SIZE_VALUE ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metrics/AllMetrics.java
khushbr: Since maxSize metric isn't collected at shard level but at node level, and `SHARD_STATS` uses `ShardID, IndexName` as dimensions, it would be confusing for the end users if this metric was added under `SHARD_STATS`.

I will rename `CACHE_CUSTOM` to `CACHE_CONFIG`.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/metrics/Cache_Max_Size.java
khushbr: Updating to use `.toString() ` instead.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metrics/AllMetrics.java
khushbr: Agree, updating this to `CACHE_MAX_SIZE_VALUE`

khushbr:
Addressing the comments in https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/pull/297

PR #274, opened 2020-07-14
Make RCA framework NOT use ClusterDetailsEventProcessor

yojs:
Using the static classes have problems as one instance of JVM can have only one instance of the class running. ClusterDetailsEventProcessor is an important class as it contains the instance details. If we want to simutate multiple RCAControllers running as part of a single JVM for the purpose of IntegrationTest framework, we should make ClusterDetailsEventProcessor a non-static class, such that each RcaController can work with its own. This is not the complete removal of the static methods in the ClusterDetailsEventProcessor, but just making the RCA framework not use them. There will be follow up PRs that will eventually make ClusterDetailsEventProcessor non-static

*Issue #, if available:*

*Description of changes:*

*Tests:*

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
@@ -92,9 +92,11 @@ public class AppContext {
   }
 
   public List<String> getPeerInstanceIps() {
-    return getAllClusterInstances().stream()
-        .skip(1)  // Skipping the first instance as it is self.
-        .map(InstanceDetails::getInstanceIp)
-        .collect(Collectors.toList());
+    return ImmutableList.copyOf(
+        getAllClusterInstances()
+            .stream()
+            .skip(1)  // Skipping the first instance as it is self.
+            .map(InstanceDetails::getInstanceIp)
+            .collect(Collectors.toList()));
   }
 }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
@@ -18,6 +18,10 @@ package com.amazon.opendistro.elasticsearch.performanceanalyzer;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.metrics.AllMetrics;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.util.InstanceDetails;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.reader.ClusterDetailsEventProcessor;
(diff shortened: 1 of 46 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/Node.java
```diff
@@ -187,4 +194,20 @@ public void setLocalFlowUnit(T localFlowUnit) {
   public void readRcaConf(RcaConf conf) {
     return;
   }
+
+  public void setAppContext(final AppContext appContext) {
+    this.appContext = appContext;
+  }
+
+  public InstanceDetails getInstanceDetails() {
+    return this.appContext.getMyInstanceDetails();
```
rguo-aws, line 203: do wen need to check whether appContext is null here to avoid null pointer exception ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/Node.java
```diff
@@ -187,4 +194,20 @@ public void setLocalFlowUnit(T localFlowUnit) {
   public void readRcaConf(RcaConf conf) {
     return;
   }
+
+  public void setAppContext(final AppContext appContext) {
```
rguo-aws, line 198: add javadoc for this function ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/Node.java
yojs: Thanks a lot @rguo-aws Will add it

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
```diff
@@ -0,0 +1,95 @@
...
+      instances.add(instanceDetails);
+    }
+    return ImmutableList.copyOf(instances);
+  }
+
+  public List<String> getPeerInstanceIps() {
+    return getAllClusterInstances().stream()
+        .skip(1)  // Skipping the first instance as it is self.
+        .map(InstanceDetails::getInstanceIp)
+        .collect(Collectors.toList());
```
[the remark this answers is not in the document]
yojs: Sure ! will add

yojs:
Good point ! removed all unused imports

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/Node.java
yojs: Do we really want a javadoc for a setter ?

PR #275, opened 2020-07-15
Fix bugs in Decision Maker

rguo-aws:
*Issue #, if available:*
#285 

*Description of changes:*
fix bugs in Decision Maker

*Tests:*
tested on docker


rguo-aws:
> Thanks for the changes @rguo-aws
> 
> I think we should do the client piece differently. Instead of having a `client` and an `rpcclient` we should have an `DecisionListener` that can be registered with the decision maker. Publisher will call this with different actions. The listener in turn can either log the action, make rpc calls, or push actions to a queue.
> 
> I can make the changes to add a listener and register it with the framework, and we can extend it to rpc client separately.
> 
> Can you please split this PR into two? One with just the bug fixes for now, and a separate one with client changes (if any). For the second PR, I suggest we use the listener approach; I'll work on a PR for that.

Thanks @vigyasharma . I have removed the client pieces and updated this PR.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decision.java
```diff
@@ -54,6 +54,11 @@ public void setDecider(String decider) {
     this.decider = decider;
   }
 
+  @Override
+  public String toString() {
+    return decider + actions;
```
vigyasharma, line 59: nit: `decider + ": " + actions` ? (assuming this is for logging?)

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
```diff
@@ -43,18 +46,30 @@ public EmptyFlowUnit operate() {
...
   public void generateFlowUnitListFromLocal(FlowUnitOperationArgWrapper args) {
-    assert true;
+    LOG.debug("Publisher: Executing fromLocal: {}", name());
+    long startTime = System.currentTimeMillis();
+
+    try {
+      this.operate();
+    } catch (Exception ex) {
+      LOG.error("Publisher: Exception in operate", ex);
+      PerformanceAnalyzerApp.ERRORS_AND_EXCEPTIONS_AGGREGATOR.updateStat(
```
khushbr, line 61: Wondering if we should update the metric count before the logging, thoughts?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decision.java
rguo-aws: done

PR #282, opened 2020-07-16
Fix PersistFlowUnitAndSummaryTest

rguo-aws:
*Issue #, if available:*
#284 

*Description of changes:*
Fix PersistFlowUnitAndSummaryTest in UTs

We found three issue when debugging this UT:
1. db contention. need to run those two UT tests in sequential order
2. DataAcessExecption is not handled  when persistor trys to create a new table 
3. the graph defined in UT was wrong. the node tag need to be changed with MASTER_DATA

*Tests:*

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PersistorBase.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PersistorBase.java
@@ -37,6 +37,7 @@ import java.util.concurrent.TimeUnit;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.jooq.Field;
+import org.jooq.exception.DataAccessException;
 
 // TODO: Scheme to rotate the current file and garbage collect older files.
 public abstract class PersistorBase implements Persistable {
@@ -211,7 +212,7 @@ public abstract class PersistorBase implements Persistable {
       T flowUnit, String tableName) throws SQLException, IOException {
     try {
         tryWriteFlowUnit(flowUnit, tableName);
-    } catch (SQLException e) {
+    } catch (SQLException | DataAccessException e) {
       LOG.info(
           "RCA: Fail to write to table '{}', try creating a new DB", tableName);
       rotateRegisterGarbageThenCreateNewDB(RotationType.FORCE_ROTATE);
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PersistorBase.java
```diff
@@ -211,7 +212,7 @@ private void rotateRegisterGarbageThenCreateNewDB(RotationType type) throws IOEx
       T flowUnit, String tableName) throws SQLException, IOException {
     try {
         tryWriteFlowUnit(flowUnit, tableName);
-    } catch (SQLException e) {
+    } catch (SQLException | DataAccessException e) {
       LOG.info(
```
yojs, line 216: Shouldn't this be `error` ?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PersistFlowUnitAndSummaryTest.java
```diff
@@ -176,35 +176,44 @@ private AppContext createAppContextWithDataNodes(String nodeName, NodeRole role)
...
     RCAScheduler scheduler = startScheduler(rcaConf, graph, persistable, this.queryable, appContext);
     // Wait at most 1 minute for the persisted data to show up with the correct contents
     WaitFor.waitFor(() -> {
       String readTableStr = persistable.read();
       System.out.println(readTableStr);
       if (readTableStr != null) {
         return readTableStr.contains("HotResourceSummary") && readTableStr.contains("DummyYoungGenRca")
-                && readTableStr.contains("HotNodeSummary") && readTableStr.contains("HotNodeRcaX")
-                && readTableStr.contains("HighHeapUsageClusterRcaX");
+                && readTableStr.contains("HotNodeSummary") && readTableStr.contains("HotNodeRcaX");
```
yojs, line 206: I see that `HighHeapUsageClusterRcaX` node is created in the `construct` but it is removed from verification here. Can you add a code-comment with the reason ?

yojs:
Please create an issue with the same tag and link it with the PR.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PersistorBase.java
rguo-aws: updated the error log

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PersistFlowUnitAndSummaryTest.java
rguo-aws: done

Issue #288, opened 2020-07-17
Read and consume decider - action priorities from yml


aditjind:
Allow customers to provide a their preferred priority order for actions mapped to a decider via a yml file and consume user provided action priorities in each decider.

PR #289, opened 2020-07-17
Reading Decider Action Priorities from YML

aditjind:
*Issue #:* #288 

*Description of changes:* Provide a list of action priorities for each decider. Read from yml and consume it in each decider

*Tests:* Manual testing in progress.


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciderConfig.yml
```diff
@@ -0,0 +1,12 @@
+# Order of action priorities
+# Loaded to a map of Decider and action priorities
+# action priorities are given as a list
+# priority order is decreasing in nature.
+---
+deciderActionPriorityOrder:
+  QueueHealthDecider:
+    - ModifyQueueCapacityAction
```
vigyasharma, line 8: We want to use decider and action names (`NAME` field, also returned by `name()`) instead of class names. This lets us refer to these names directly at different places in the code.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
```diff
@@ -0,0 +1,22 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class ConfigReader {
+    public static HashMap<Decider, ArrayList<String>> getDeciderActionPriorityOrder() {
+        ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+        HashMap<Decider, ArrayList<String>> deciderActionPriorityOrder = null;
```
vigyasharma, line 13: Looks like the yml is read and created each time someone fetches action priority. How about parsing and storing it ones, during load/init?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -81,8 +81,8 @@ public Decision operate() {
   }
 
   private void configureActionPriority() {
-    // TODO: Input from user configured yml
-    this.actionsByUserPriority.add(ModifyQueueCapacityAction.NAME);
+    ArrayList<String> actionPriority = getDeciderActionPriorityOrder(this.getClass().getName());
```
vigyasharma, line 84: Use `NAME` instead of `this.getClass().getName()`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decider.java
```diff
@@ -41,10 +45,12 @@
 
   private static final Logger LOG = LogManager.getLogger(Decider.class);
   protected final int decisionFrequency; // Measured in terms of number of evaluationIntervalPeriods
+  private final HashMap<Decider, ArrayList<String>> deciderActionPriorityOrder;
 
   public Decider(long evalIntervalSeconds, int decisionFrequency) {
     super(0, evalIntervalSeconds);
     this.decisionFrequency = decisionFrequency;
+    this.deciderActionPriorityOrder = ConfigReader.getDeciderActionPriorityOrder();
```
vigyasharma, line 53: If `ConfigReader` can store the priority hashmap in memory, then each decider doesn't need to store it. Config reader can expose the getter methods

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
```diff
@@ -0,0 +1,22 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker;
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.deciders.Decider;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class ConfigReader {
```
vigyasharma, line 10: nit: rename to something specific to action priorities?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciderConfig.yml
```diff
@@ -0,0 +1,12 @@
+# Order of action priorities
```
vigyasharma, line 1: Where will this yml be present on a deployed cluster? It cant be within the jar, users need to be able to replace it without building jars. Is that possible with current implementation? I think we should also have a path configuration which can be passed as jvm args in case users need to change its default location.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
```diff
@@ -0,0 +1,22 @@
...
+    public static HashMap<Decider, ArrayList<String>> getDeciderActionPriorityOrder() {
+        ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
```
rguo-aws, line 12: Is there any specific reason to from from YML instead of JSON ? All RCA related configs are stored in JSON file and why do we want to switch to a different format in decision maker ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
```diff
@@ -0,0 +1,22 @@
...
+        HashMap<Decider, ArrayList<String>> deciderActionPriorityOrder = null;
+        try {
+            deciderActionPriorityOrder = mapper.readValue(new File("deciderConfig.yml"), HashMap.class);
+            return deciderActionPriorityOrder;
+        } catch (Exception e) {
+            e.printStackTrace();
```
rguo-aws, line 18: Add an error  log here ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class ConfigReader {
+    public static HashMap<Decider, ArrayList<String>> getDeciderActionPriorityOrder() {
+        ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+        HashMap<Decider, ArrayList<String>> deciderActionPriorityOrder = null;
+        try {
+            deciderActionPriorityOrder = mapper.readValue(new File("deciderConfig.yml"), HashMap.class);
+            return deciderActionPriorityOrder;
```
rguo-aws, line 16: this return statement seems to be redundant

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.deciders.Decider;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class ConfigReader {
+    public static HashMap<Decider, ArrayList<String>> getDeciderActionPriorityOrder() {
```
rguo-aws, line 11: Can we add Nullable annotation to this function ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
aditjind: Done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decider.java
aditjind: Modified in the latest version.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciderConfig.yml
aditjind: Gotcha. Updated.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
aditjind: Updated

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
aditjind: Yup. Updated in the latest version.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciderConfig.yml
aditjind: I have modified the location to be in the same repo as the other `PA_CONFIGs`. I am thinking that this file can be thought of as a one stop config for all the decider related configs. Do we really want to give the knob of keeping all of them modifiable by the customer? Also, what benefit does changing the location brings? The action priority should be a one time setup from the customer during the cluster setup based on the type of the load customer anticipates and tuning after the setup. Once the initial setup is done Do we still anticipate these priorities changing?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class ConfigReader {
+    private static final Logger LOG = LogManager.getLogger(ConfigReader.class);
+    private static HashMap<Decider, ArrayList<String>> deciderActionPriorityOrder;
+
+    public static final String CONFIG_FILE_PATH = "./pa_config/deciderConfig.yml";
```
vigyasharma, line 16: Is this is the right way to provide file path for configuration files? What is the current directory here for `./...`?
Please see how path has been specified for rca.conf in `RcaConsts.java` and change accordingly.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public ConfigReader() {
```
vigyasharma, line 18: The constructor can take path as input to support test cases with different config files.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
vigyasharma: I think we should limit this one to only action priorities, to keep it simple to read. Configuration values could go into rca.conf along with other config values. But you can take a call on that if you think there are other decider level configs that should belong here. 
Also, if you want to make it generic you (at least) need another level of hierarchy in the yml, right?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
```diff
@@ -0,0 +1,36 @@
...
+        this.deciderActionPriorityOrder = null;
+    }
+
+    public static void updateDeciderActionPriorityOrder() {
+        ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+        try {
+            deciderActionPriorityOrder = mapper.readValue(new File(CONFIG_FILE_PATH), HashMap.class);
```
vigyasharma, line 25: This will also not work with the *generic* decider config file you mentioned. Let us just restrict it to an action_priorities file only

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/ConfigReader.java
```diff
@@ -0,0 +1,36 @@
...
+        } catch (Exception e) {
+            LOG.error("Could not read the Decider Config File");
+            e.printStackTrace();
+        }
+        return;
+    }
+
+    public static ArrayList<String> getActionPriorityOrder(String decider) {
+        return deciderActionPriorityOrder.get(decider);
```
vigyasharma, line 34: Handle NPE here?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -81,8 +80,8 @@ public Decision operate() {
   }
 
   private void configureActionPriority() {
-    // TODO: Input from user configured yml
-    this.actionsByUserPriority.add(ModifyQueueCapacityAction.NAME);
+    ArrayList<String> actionPriority = DeciderActionPriorityReader.getActionPriorityOrder(name());
+    actionsByUserPriority = actionPriority;
```
vigyasharma, line 84: nit: we can merge these lines

rguo-aws: +1.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElasticSearchAnalysisGraph.java
```diff
@@ -175,6 +177,10 @@ public void construct() {
     queueRejectionClusterRca.addAllUpstreams(Collections.singletonList(queueRejectionNodeRca));
     queueRejectionClusterRca.addTag(TAG_AGGREGATE_UPSTREAM, LOCUS_DATA_NODE);
 
+    // Upload the config from YAML into memory.
+    DeciderActionPriorityReader reader = new DeciderActionPriorityReader(DECIDER_ACTION_PRIORITIES_PATH);
```
rguo-aws, line 181: let's try to move this reader to somewhere else as it is clearly not part of the RCA graph.

PR #291, opened 2020-07-20
More refactoring to plugin-in an integration test framework for RCA components

yojs:
*Issue #:* #292

*Description of changes:*

The changes are mostly to expose interfaces that were otherwise buried into private interfaces.
It also removes some of the static usages in the RcaController
Changes a few tests accordingly to account for the changes

*Tests:*
No new tests are required as this is just refactoring. Existing tests should cover the changes.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/NetServer.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/NetServer.java
@@ -89,7 +89,7 @@ public class NetServer extends InterNodeRpcServiceGrpc.InterNodeRpcServiceImplBa
    */
   protected Server server;
 
-  private boolean attemptedShutdown;
+  private volatile boolean attemptedShutdown;
 
   public NetServer(final int port, final int numServerThreads, final boolean useHttps) {
     this.port = port;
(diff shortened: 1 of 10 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -114,6 +115,8 @@
   private final long rcaStateCheckIntervalMillis;
   private final long roleCheckPeriodicity;
 
+  private boolean deliberateInterrupt;
```
rguo-aws, line 118: does this need to be declared as volatile ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/NetServer.java
```diff
@@ -89,10 +89,13 @@
    */
   protected Server server;
 
+  private boolean attemptedShutdown;
```
rguo-aws, line 92: declare this as volatile ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -251,8 +285,10 @@ public void run() {
           Thread.sleep(rcaStateCheckIntervalMillis - duration);
         }
       } catch (InterruptedException ie) {
-        LOG.error("RCA controller thread was interrupted. Reason: {}", ie.getMessage());
-        LOG.error(ie);
+        if (!deliberateInterrupt) {
```
ktkrg, line 288: Looks like this was missed in the original code as well, but we should also propagate the interrupted exception if we caught it by calling Thread.currentThread().interrupt() explicitly here.

PR #294, opened 2020-07-20
Common changes needed to support dynamic en/disabling of config overrides

### LINKED ISSUE #290: Support dynamic enable/disabling of RCA, decider, and action nodes
This issue captures the changes that are needed in the reader to support writing additional lines in the ClusterDetailsEvents section and the common code that does not depend on Elasticsearch classes.

This is a prerequisite for opendistro-for-elasticsearch/performance-analyzer#147

ktkrg:
*Issue #:*
Fixes #290 
*Description of changes:*
This change adds the common code needed  for supporting dynamic enabling and disabling of rcas, deciders, and actions.

The actual changes for applying the config overrides will be in the next PR after this gets merged. It's currently in a branch: performance-analyzer-rca/ktkrg-ed2 branch.

*Tests:*


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesHelper.java
```diff
@@ -0,0 +1,65 @@
...
+      try {
+        return MAPPER.readValue(overrides, ConfigOverrides.class);
+      } catch (IOException ioe) {
+        exception[0] = ioe;
+      }
+      return null;
+    });
+
+    if (configOverrides == null && exception[0] != null) {
+      // re throw the exception that was consumed while deserializing.
```
khushbr, line 59: Do we also want to add a metric here ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverrides.java
```diff
@@ -0,0 +1,69 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.config.overrides;
```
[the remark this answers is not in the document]
ktkrg: Thanks for noticing it! Added them.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverrides.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public static class Overrides {
+        private List<String> rcas;
+        private List<String> deciders;
+        private List<String> actions;
+
+        public Overrides() {
+            this.rcas = new ArrayList<>();
+            this.deciders = new ArrayList<>();
+            this.actions = new ArrayList<>();
```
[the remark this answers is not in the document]
ktkrg: Added javadoc.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesHelper.java
ktkrg: I've added the metric where we call deserialize() in ClusterDetailsEventProcessor, I'll send the PR after this gets merged as that PR depends on this change.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesHelper.java
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * Deserializes a JSON representation of the config overrides into a {@link ConfigOverrides} instance.
+   *
+   * @param overrides The JSON string representing config overrides.
+   * @return A {@link ConfigOverrides} instance if the JSON is valid.
+   * @throws IOException if conversion runs into an IOException.
+   */
+  public static ConfigOverrides deserialize(final String overrides) throws IOException {
+    final IOException[] exception = new IOException[1];
```
yojs, line 63: Can you explain if this is a one element array, why does it needs to be an array, can't it just be a single reference ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesHelper.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.IOException;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+/**
+ * Class that helps with operations concerning {@link ConfigOverrides}s
+ */
+public class ConfigOverridesHelper {
+
+  private static final ObjectMapper MAPPER = new ObjectMapper();
```
yojs, line 28: This is a static final object. Some of the decider actions are node level. I don't think this will work with the RCA-IT framework if we just have a static object mapper here, unless the Overrides object haves nodeId as a parameter, in my opinion.

ktkrg: This is only to help deserialize the string from ClusterDetailsEventProcessor into the override object. It's then updated in the RcaConf. This change will come in the next PR. This PR is only introducing the common classes needed by both the reader and the writer.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesHelper.java
ktkrg: We can't use a local variable inside the lambda because it needs to be effectively final so we need a one element array(or any wrapper around the exception) to fish the exception out of the lambda into the calling code. I've added this as a javadoc comment as well.

PR #295, opened 2020-07-20
Add listeners for publisher actions

vigyasharma:
*Issue #:* 316
https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/issues/316

*Description of changes:* Allows users to plugin different action listeners to publisher. Action suggestions are published to each listener.

*Tests:* WIP

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
@@ -28,10 +28,10 @@ import com.google.common.annotations.VisibleForTesting;
 
 import java.time.Instant;
 import java.util.ArrayList;
-import java.util.concurrent.TimeUnit;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
@@ -17,6 +17,7 @@ package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.de
 
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.PerformanceAnalyzerApp;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.Action;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.ActionListener;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.FlipFlopDetector;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.TimedFlipFlopDetector;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.core.NonLeafNode;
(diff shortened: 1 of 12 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
```diff
@@ -30,10 +33,12 @@
 
   private Collator collator;
   private boolean isMuted = false;
+  private List<ActionListener> actionListeners;
```
rguo-aws, line 36: should we make this list thread safe ? it looks to me that addActionListener() is called from a different thread.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/plugins/PluginController.java
```diff
@@ -0,0 +1,72 @@
...
+      if (constructors[0].getParameterCount() != 0) {
+        throw new IllegalStateException(
+            "default constructor expected for plugin class: [" + pluginClass.getName() + "]");
+      }
+
+      try {
+        plugins.add((Plugin) constructors[0].newInstance());
+        LOG.info("loaded plugin: [{}]", plugins.get(plugins.size() - 1).name());
+      } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
+        e.printStackTrace();
```
rguo-aws, line 60: maybe we should add an error log here ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/plugins/PluginController.java
```diff
@@ -0,0 +1,72 @@
...
+public class PluginController {
+
+  private static final Logger LOG = LogManager.getLogger(PluginController.class);
+  private final Publisher publisher;
+  private List<Plugin> plugins;
+
+  public PluginController(Publisher publisher) {
+    this.publisher = publisher;
+    this.plugins = new ArrayList<>();
+    loadFrameworkPlugins();
```
rguo-aws, line 36: IllegalStateException is not handled here.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/plugins/PluginController.java
sidheart: ^ seconded can we replace this with Log.error("message", e)? That'll include the stacktrace.

vigyasharma: Added error log here and throwing a runtime exception. Since these are currently framework plugins, which are critical, I want the process to error out instead of silently logging them.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/plugins/PluginController.java
vigyasharma: We want the system to error out as plugins can be critical.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
vigyasharma: It should only be called during instantiation of the graph, not during regular scheduling.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/plugins/PluginControllerConfig.java
```diff
@@ -0,0 +1,31 @@
...
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.plugins;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PluginControllerConfig {
```
sidheart, line 21: Can we use a singleton pattern here instead of static? Then we can avoid the PowerMock stuff in the test class and have multiple configs in one JVM context

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/plugins/PluginControllerTest.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.List;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mockito;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PowerMockIgnore;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+@PowerMockIgnore({"com.sun.org.apache.xerces.*", "org.apache.logging.log4j.*"})
```
sidheart, line 36: I have a strong distaste for PowerMock, it breaks code coverage analysis since it generates new bytecode and it causes a host of compiler warning that will eventually block us from building our code with newer versions of java. Can we use a singleton pattern for PluginControllerConfig and mock it using Mockito instead?

PR #298, opened 2020-07-22
Add a cluster level collector for node config settings

rguo-aws:
*Issue #:*
#300 

*Description of changes:*
1. Add a cluster level collector for node config settings
2. create a thread-safe node config cache in AppContect to store node config settings and share them among different RCA vertices in RCA graph

*Tests:*
tested on docker

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCache.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCache.java
@@ -30,25 +30,38 @@ import org.apache.commons.lang3.builder.HashCodeBuilder;
  */
 public class NodeConfigCache {
 
-  private static final int LOADING_CACHE_EVICTION_TIMEOUT = 10;
+  private static final int CACHE_TTL = 10;
   private final Cache<NodeConfigKey, Double> nodeConfigCache;
 
-  //unbounded cache with eviction timeout set to 10 mins
+  //unbounded cache with TTL set to 10 mins
   public NodeConfigCache() {
     nodeConfigCache =
         CacheBuilder.newBuilder()
-            .expireAfterWrite(LOADING_CACHE_EVICTION_TIMEOUT, TimeUnit.MINUTES)
+            .expireAfterWrite(CACHE_TTL, TimeUnit.MINUTES)
             .build();
   }
 
+  /**
+   * add config value into cache
+   * @param nodeKey the NodeKey of the node on which this config is collected
+   * @param config the config type
+   * @param value the config value
+   */
   public void put(NodeKey nodeKey, Resource config, double value) {
     nodeConfigCache.put(new NodeConfigKey(nodeKey, config), value);
   }
 
-  public double get(NodeKey nodeKey, Resource config) {
+  /**
+   * returns the config value that is associated with the nodeKey and config
+   * @param nodeKey the NodeKey of the node
+   * @param config the config type
+   * @return the config value
+   * @throws IllegalArgumentException throws an exception if the config does not exist in cache
+   */
+  public double get(NodeKey nodeKey, Resource config) throws IllegalArgumentException {
     Double ret = nodeConfigCache.getIfPresent(new NodeConfigKey(nodeKey, config));
     if (ret == null) {
-      return Double.NaN;
+      throw new IllegalArgumentException();
     }
     return ret;
   }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCache.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCache.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
 package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.collector;
 
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.Resource;
@@ -8,9 +23,10 @@ import java.util.concurrent.TimeUnit;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 
 /**
- * we create a cache instance in {@link com.amazon.opendistro.elasticsearch.performanceanalyzer.AppContext}
- * to store the node config settings from each node. Key is NodeKey + Resource and value is the
- * actual value of the config setting(i.e. size of write queue capacity)
+ * we create a thread-safe unbounded cache instance in {@link com.amazon.opendistro.elasticsearch.performanceanalyzer.AppContext}
+ * to store the node config settings from each node. Any RCA vertex in RCA graph can read the node config directly from
+ * this cache instance. The key of this cache is NodeKey + Resource and value is the actual value of the config setting
+ * (i.e. size of write queue capacity)
  */
 public class NodeConfigCache {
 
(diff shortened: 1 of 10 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCache.java
```diff
@@ -0,0 +1,90 @@
...
+
+  public void put(NodeKey nodeKey, Resource config, double value) {
+    nodeConfigCache.put(new NodeConfigKey(nodeKey, config), value);
+  }
+
+  public double get(NodeKey nodeKey, Resource config) {
+    Double ret = nodeConfigCache.getIfPresent(new NodeConfigKey(nodeKey, config));
+    if (ret == null) {
+      return Double.NaN;
+    }
```
vigyasharma, line 52: Since this is not going to sql db, I don't think we need `NaN` value and check. Let's instead throw a checked exception here for consumer code safety? (anyway the check has to be handled by the caller; exception will help protect against cases where callers forget the `NaN` check).

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigClusterCollector.java
```diff
@@ -0,0 +1,105 @@
...
+ * Cluster level node config collector that collect node config settings from each node and
+ * store them into the {@link NodeConfigCache}
+ */
+public class NodeConfigClusterCollector extends NonLeafNode<EmptyFlowUnit> {
+
+  private static final Logger LOG = LogManager.getLogger(NodeConfigClusterCollector.class);
+  private final NodeConfigCollector nodeConfigCollector;
+
+  public NodeConfigClusterCollector(final NodeConfigCollector nodeConfigCollector) {
+    super(0, 5);
```
khushbr, line 41: The constants here make the code unreadable, if we can add a variable with name suggesting what the constant represents, it is helpful. 

Additionally, later we should refactor to introduce a defaultIntervalPeriod and use is across all collector and similarly for RCAs as well.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigClusterCollector.java
```diff
@@ -0,0 +1,105 @@
...
+    this.nodeConfigCollector = nodeConfigCollector;
+  }
+
+  private void addNodeLevelConfigs() {
```
khushbr, line 45: nit:  Javadoc explaining what `addNodeLevelConfigs()` is doing.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCache.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+public class NodeConfigCache {
+
+  private static final int LOADING_CACHE_EVICTION_TIMEOUT = 10;
+  private final Cache<NodeConfigKey, Double> nodeConfigCache;
+
+  //unbounded cache with eviction timeout set to 10 mins
+  public NodeConfigCache() {
+    nodeConfigCache =
+        CacheBuilder.newBuilder()
```
aditjind, line 39: Just Curious: Any particular reason we picked the unbounded cache? We may not be having fixed number of entries we want to put in the cache. So wouldn't leaving it unbounded might cause out of memory errors? It should be fine, If we are sure the number of entries are less.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/store/collector/NodeConfigCacheTest.java
```diff
@@ -0,0 +1,64 @@
...
+    nodeConfigCache.put(nodeKey1, ResourceUtil.WRITE_QUEUE_CAPACITY, 2.0);
+    val = nodeConfigCache.get(nodeKey1, ResourceUtil.WRITE_QUEUE_REJECTION);
+    Assert.assertTrue(Double.isNaN(val));
+
+    val = nodeConfigCache.get(nodeKey2, ResourceUtil.WRITE_QUEUE_CAPACITY);
+    Assert.assertTrue(Double.isNaN(val));
+  }
+
+  @Test
+  public void testSetAndGetValue() {
```
aditjind, line 55: Maybe 1 UT for updating the value of the key and asserting if the updated value is returned.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCache.java
rguo-aws: done, throw an exception if config does not exist in cache

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCache.java
```diff
@@ -0,0 +1,90 @@
...
+            .expireAfterWrite(LOADING_CACHE_EVICTION_TIMEOUT, TimeUnit.MINUTES)
+            .build();
+  }
+
+  public void put(NodeKey nodeKey, Resource config, double value) {
+    nodeConfigCache.put(new NodeConfigKey(nodeKey, config), value);
```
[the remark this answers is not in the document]
rguo-aws: we perform this check the NodeConfigCollector when collection configs from DB.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCache.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * we create a thread-safe unbounded cache instance in {@link com.amazon.opendistro.elasticsearch.performanceanalyzer.AppContext}
+ * to store the node config settings from each node. Any RCA vertex in RCA graph can read the node config directly from
+ * this cache instance. The key of this cache is NodeKey + Resource and value is the actual value of the config setting
+ * (i.e. size of write queue capacity)
+ */
+public class NodeConfigCache {
+
+  private static final int LOADING_CACHE_EVICTION_TIMEOUT = 10;
```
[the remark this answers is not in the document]
rguo-aws: done, replaced it with TTL

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCache.java
rguo-aws: the maxmium number of entries will be (number of node) * (node config on each node) which is not that big considering that stale entries will be evicted if reaches TTL.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCache.java
```diff
@@ -0,0 +1,90 @@
...
+            .expireAfterWrite(LOADING_CACHE_EVICTION_TIMEOUT, TimeUnit.MINUTES)
+            .build();
+  }
+
+  public void put(NodeKey nodeKey, Resource config, double value) {
+    nodeConfigCache.put(new NodeConfigKey(nodeKey, config), value);
+  }
+
+  public double get(NodeKey nodeKey, Resource config) {
+    Double ret = nodeConfigCache.getIfPresent(new NodeConfigKey(nodeKey, config));
```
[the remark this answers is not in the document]
rguo-aws: we don't handle stale data check in the cache itself. data will be return if presented in cache and we will evict stale entries only if TTL is reaches

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigClusterCollector.java
rguo-aws: yes, we can create a static const here.  But since we will do another refactoring and remove the 5 second intervals form the super() constructor completely, it would be better to leave it as is at this moment ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigClusterCollector.java
rguo-aws: done. javadoc added

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigClusterCollector.java
khushbr: Makes sense, please add an issue for this.

PR #304, opened 2020-07-24
[bugfix] update the ClusterDetailsEventProcessor only when it is valid.

yojs:
There was a bug where we were updating the ClusterDetailsEventProcessor before it was filled with data. This fixes it and some other cases where we can run into NPE.

*Issue #:* #305

*Description of changes:* ReaderMetricsProcessor skips updating the ClusterDetails every other time this is when the AppContext will be set with a cluster details that says the current node role is unknown. This would be the reason the RCA rest API would complain about the current node not being the elected master. 

The change in the node role can also cause the RcaScheduler to not start up.

*Tests:* Tested it on docker where the error was reported.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
@@ -45,6 +45,9 @@ public class AppContext {
 
   public AppContext(AppContext other) {
     this.clusterDetailsEventProcessor = new ClusterDetailsEventProcessor(other.clusterDetailsEventProcessor);
+
+    // Initializing this as we don't want to copy the entire cache.
+    this.nodeConfigCache = new NodeConfigCache();
   }
 
   public void setClusterDetailsEventProcessor(final ClusterDetailsEventProcessor clusterDetailsEventProcessor) {
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
@@ -43,6 +43,10 @@ public class AppContext {
     this.nodeConfigCache = new NodeConfigCache();
   }
 
+  public AppContext(AppContext other) {
+    this.clusterDetailsEventProcessor = new ClusterDetailsEventProcessor(other.clusterDetailsEventProcessor);
+  }
+
   public void setClusterDetailsEventProcessor(final ClusterDetailsEventProcessor clusterDetailsEventProcessor) {
     this.clusterDetailsEventProcessor = clusterDetailsEventProcessor;
   }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
@@ -362,7 +362,7 @@ class SQLitePersistor extends PersistorBase {
         response.addNestedSummaryList(nodeLevelDimSummary);
       }
     } catch (DataAccessException dex) {
-      LOG.error("Failed to read temperature profile RCA for {}", rca, dex);
+      LOG.warn("Failed to read temperature profile RCA for {}. Msg: {}", rca, dex.getMessage());
       if (dex.getMessage().contains("no such table")) {
         JsonObject json = new JsonObject();
         json.addProperty("error", "RCAs are not created yet.");
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
@@ -362,11 +362,12 @@ class SQLitePersistor extends PersistorBase {
         response.addNestedSummaryList(nodeLevelDimSummary);
       }
     } catch (DataAccessException dex) {
-      LOG.warn("Failed to read temperature profile RCA for {}. Msg: {}", rca, dex.getMessage());
       if (dex.getMessage().contains("no such table")) {
         JsonObject json = new JsonObject();
         json.addProperty("error", "RCAs are not created yet.");
         return json;
+      } else {
+        LOG.error("Failed to read temperature profile RCA for {}.", rca, dex);
       }
     }
     return response.toJson();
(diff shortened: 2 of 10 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
```diff
@@ -362,7 +362,7 @@ private JsonElement readTemperatureProfileRca(String rca) {
         response.addNestedSummaryList(nodeLevelDimSummary);
       }
     } catch (DataAccessException dex) {
-      LOG.error("Failed to read temperature profile RCA for {}", rca, dex);
+      LOG.warn("Failed to read temperature profile RCA for {}. Msg: {}", rca, dex.getMessage());
```
sidheart, line 365: This is generally an anti pattern. Is there a reason you're throwing away the stacktrace?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
```diff
@@ -38,6 +38,10 @@ public AppContext() {
     this.clusterDetailsEventProcessor = null;
   }
 
+  public AppContext(AppContext other) {
+    this.clusterDetailsEventProcessor = new ClusterDetailsEventProcessor(other.clusterDetailsEventProcessor);
```
rguo-aws, line 42: should we also can the default constructor this() here in case something else needs to be initialized in AppContext ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
```diff
@@ -40,6 +40,19 @@
    */
   private volatile ImmutableList<NodeDetails> nodesDetails = null;
 
+  public ClusterDetailsEventProcessor() {}
+
+  public ClusterDetailsEventProcessor(final ClusterDetailsEventProcessor other) {
+    if (other.nodesDetails != null) {
```
rguo-aws, line 46: same as above

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -466,6 +463,10 @@ is ready so it starts to read that file (go back two windows and
 
     emitMetrics(currWindowStartTime);
 
+    if (appContext != null && !clusterDetailsEventsProcessor.getNodesDetails().isEmpty()) {
```
rguo-aws, line 466: should we add a few comments here to explain why we want to do this check ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
```diff
@@ -40,6 +40,19 @@
...
+      List<NodeDetails> nodeDetails = new ArrayList<>();
```
sidheart, line 47: Nit: use a builder for this instead of an alloc + copy
```
ImmutableList.Builder builder = new ImmutableList.Builder<NodeDetails>();
for (final NodeDetails oldDetails : other.nodesDetails) {
        builder.add(new NodeDetails(oldDetails));
}
this.nodesDetails = builder.build();
```

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
```diff
@@ -148,6 +161,13 @@ public NodeDetails(AllMetrics.NodeRole role, String id, String hostAddress, bool
       this.role = role.toString();
     }
 
+    public NodeDetails(NodeDetails other) {
```
sidheart: Nit: Annotate with @Nonnull or add null check

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
yojs: Not being able to find the RCA table is justified if the query is made early enough that one is not generated. The long stack trace gives people an impression of an error, when this can happen. That's why I removed the stacktrace with just a message.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
sidheart: ^ I was confused by this too, I had to look into AppContext to figure out why a comment would be nice

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
yojs: This is a copy constructor. So ideally it should copy everything but after rebasing I saw, you added a new member which is a cache. I am sure we don't want to make a copy of it. 
So, I will call `this()` as you sugested

yojs: Thinking more about it, I will explicitly initialize the other. Calling other() initializes the other with default which we may not intend.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
yojs: Similar argument

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
yojs: Good point. Will do

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
yojs: Thanks a lot ! good point

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
yojs: makes sense. Thanks !

yojs:
I am merging this PR was the spotBugs failures introduced by [this](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/runs/905025845) are being looked at by @aditjind

PR #308, opened 2020-07-25
Fixing Writing to Static Fields in Decider Action priorities

aditjind:
*Issue #:* #288 #289 

*Description of changes:* Fixing Writing to Static Fields in Decider Action priorities

*Tests:*

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
@@ -36,14 +36,13 @@ public class QueueHealthDecider extends Decider {
   public static final String NAME = "queue_health";
 
   private QueueRejectionClusterRca queueRejectionRca;
-  List<String> actionsByUserPriority = new ArrayList<>();
+  List<String> actionsByUserPriority = null;
   private int counter = 0;
 
   public QueueHealthDecider(long evalIntervalSeconds, int decisionFrequency, QueueRejectionClusterRca queueRejectionClusterRca) {
     // TODO: Also consume NodeConfigurationRca
     super(evalIntervalSeconds, decisionFrequency);
     this.queueRejectionRca = queueRejectionClusterRca;
-    configureActionPriority();
   }
 
   @Override
@@ -79,8 +78,8 @@ public class QueueHealthDecider extends Decider {
     return decision;
   }
 
-  private void configureActionPriority() {
-    actionsByUserPriority = DeciderActionPriorityReader.getActionPriorityOrder(name());
+  public void configureActionPriority(ArrayList<String> priority) {
+    actionsByUserPriority = priority;
   }
 
   /**
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElasticSearchAnalysisGraph.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElasticSearchAnalysisGraph.java
@@ -15,12 +15,14 @@
 
 package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store;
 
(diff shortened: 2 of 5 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElasticSearchAnalysisGraph.java
```diff
@@ -175,10 +177,16 @@ public void construct() {
...
+    DeciderActionPriorityReader reader = new DeciderActionPriorityReader(DECIDER_ACTION_PRIORITIES_PATH);
+    reader.updateDeciderActionPriorityOrder();
+
     // Queue Health Decider
-    QueueHealthDecider queueHealthDecider = new QueueHealthDecider(EVALUATION_INTERVAL_SECONDS, 12, queueRejectionClusterRca);
+    QueueHealthDecider queueHealthDecider = new QueueHealthDecider(EVALUATION_INTERVAL_SECONDS, 12,
+            queueRejectionClusterRca);
     queueHealthDecider.addTag(TAG_LOCUS, LOCUS_MASTER_NODE);
     queueHealthDecider.addAllUpstreams(Collections.singletonList(queueRejectionClusterRca));
+    queueHealthDecider.configureActionPriority(reader.getActionPriorityOrder(queueHealthDecider.name()));
```
vigyasharma, line 189: *Ideally*, I think `deciders` should have control of when/how their action priorities are populated. As such, the priority config reader should just be passed to each decider, to use as they see fit. In this current form, it also makes it harder to refactor/split creation of these deciders, which I feel is a bigger problem. DI (Guice) can probably help here but I'm happy to consider other ideas around this too.

Also, since currently the yml is read during decider creation, it will never get refreshed without a process restart, right? This *may* be okay, although it is better if we can write it in a way that doesn't close the doors on refreshing config w/o process restarts. Right now (and in foreseeable future) we are not doing anything in deciders that should need a process restart (like opening/closing resources that need to be purged/changed).

aditjind: Deciders can have control on when specific priorities to them are populated from the in-memory population of all the priorities. I don't know of a way (if one exists) where we can read specific chunks of the same config from different parts. If the way doesn't exist then we will need to have a way to periodically refresh the whole config or give the refreshing flexibility to decider for each decider specific config. The solution here can be to split the config into different decider files altogether and have an interface implementation of the reader. But having an interface for just reading the config from a file seems like an overkill. 

The problem with the initial implementation was that the static values (the file path and the config) were being updated from a non static context (constructors). This is not a good code practice and failed the SpotBugs build. This PR has been raised as an extension to the original one.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -36,14 +36,13 @@
...
   private QueueRejectionClusterRca queueRejectionRca;
-  List<String> actionsByUserPriority = new ArrayList<>();
+  List<String> actionsByUserPriority = null;
   private int counter = 0;
 
   public QueueHealthDecider(long evalIntervalSeconds, int decisionFrequency, QueueRejectionClusterRca queueRejectionClusterRca) {
     // TODO: Also consume NodeConfigurationRca
     super(evalIntervalSeconds, decisionFrequency);
     this.queueRejectionRca = queueRejectionClusterRca;
-    configureActionPriority();
```
sidheart, line 46: Shouldn't this be replaced by a call to configureActionPriority(DeciderActionPriorityReader.getActionPriorityOrder(name()))?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElasticSearchAnalysisGraph.java
vigyasharma: Deciders don't need to read from offsets in the yml, they can get the entire configuration hashmap and pick values relevant to them. What I meant was that we should decouple instantiation of deciders from setting action priorities for deciders.

Issue #309, opened 2020-07-27
GRPC on any port


yojs:
Today the networking layer assumes that on the remote node, gRPC is always running on the configured port, or to put it differently, all nodes in the cluster are running their gRPC servers on the same port. This causes a problem for the RCA Integration Test (IT) framework. in RCA IT, we simulate multiple nodes of the cluster running as different threads on the same machine. So, for the IT framework, all gRPC servers have the same IP addresses but different ports. Therefore, it is important that we support this

PR #310, opened 2020-07-27
Grpc server can run on different ports on different nodes of the cluster

yojs:
*Issue #:* #309 

*Description of changes:*
- The gRPC port is determined by the new field in the `InstanceDetails` class.
- Now gRPC server can run on different ports on different nodes of the cluster. 
- Use `node ID` as the primary way of distinguishing nodes in the networking layer rather than the IP addresses, for cases when all nodes run on the same machine, they will have the same IP but node IDs will still be unique.
- Because `ID` and `IP` are both represented as strings, it wasn't possible to distinguish them in the method arguments and one can slip in one for the other until the gRPC server creation complains. This PR makes them type safe by wrapping them in class of their own as java does not provide type aliasing.

*Tests:*
New tests are not added. All existing tests should continue to run correctly.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
@@ -115,7 +115,8 @@ public class PluginSettings {
   }
 
   @VisibleForTesting
-  public @Nullable String getProperty(String key) {
+  @Nullable
+  public String getProperty(String key) {
     return settings.getProperty(key);
   }
 
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
@@ -22,6 +22,7 @@ import com.google.common.annotations.VisibleForTesting;
 import java.io.File;
 import java.util.Properties;
 
+import javax.annotation.Nullable;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.message.ParameterizedMessage;
@@ -111,6 +112,11 @@ public class PluginSettings {
   @VisibleForTesting
   public void overrideProperty(String key, String value) {
     settings.setProperty(key, value);
+  }
+
+  @VisibleForTesting
+  public @Nullable String getProperty(String key) {
+    return settings.getProperty(key);
   }
 
   public boolean shouldCleanupMetricsDBFiles() {
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decision.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decision.java
@@ -19,6 +19,7 @@ import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.act
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.FlowUnitMessage;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.core.GenericFlowUnit;
 
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.util.InstanceDetails;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -65,7 +66,7 @@ public class Decision extends GenericFlowUnit {
   }
 
   @Override
-  public FlowUnitMessage buildFlowUnitMessage(String graphNode, String esNode) {
+  public FlowUnitMessage buildFlowUnitMessage(String graphNode, InstanceDetails.Id esNode) {
     // All deciders run on the master node, (in initial versions), so we dont expect Decisions
     // to be passed over wire.
     throw new IllegalStateException(
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
@@ -19,6 +19,7 @@ import com.amazon.opendistro.elasticsearch.performanceanalyzer.CertificateUtils;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.core.Util;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.InterNodeRpcServiceGrpc;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.InterNodeRpcServiceGrpc.InterNodeRpcServiceStub;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.util.InstanceDetails;
 import com.google.common.annotations.VisibleForTesting;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
@@ -54,17 +55,15 @@ public class GRPCConnectionManager {
   private File trustedCasFile;
 
   /**
-   * Map of remote host to a Netty channel to that host.
+   * Map of remote hostId to a Netty channel to that host.
    */
-  private ConcurrentMap<String, AtomicReference<ManagedChannel>> perHostChannelMap =
-      new ConcurrentHashMap<>();
+  private ConcurrentMap<InstanceDetails.Id, AtomicReference<ManagedChannel>> perHostChannelMap = new ConcurrentHashMap<>();
 
   /**
-   * Map of remote host to a grpc client object of that host. The client objects are created over
+   * Map of remote hostId to a grpc client object of that host. The client objects are created over
    * the channels for those hosts and are used to call RPC methods on the hosts.
    */
-  private ConcurrentMap<String, AtomicReference<InterNodeRpcServiceStub>> perHostClientStubMap =
-      new ConcurrentHashMap<>();
+  private ConcurrentMap<InstanceDetails.Id, AtomicReference<InterNodeRpcServiceStub>> perHostClientStubMap = new ConcurrentHashMap<>();
 
   /**
    * Flag that controls if we need to use a secure or an insecure channel.
@@ -73,7 +72,7 @@ public class GRPCConnectionManager {
 
   public GRPCConnectionManager(final boolean shouldUseHttps) {
     this.shouldUseHttps = shouldUseHttps;
-    this.port = Util.RPC_PORT;
+    this.port = 0;
     if (shouldUseHttps) {
       this.certFile = CertificateUtils.getClientCertificateFile();
       this.pkeyFile = CertificateUtils.getClientPrivateKeyFile();
@@ -97,12 +96,12 @@ public class GRPCConnectionManager {
   }
 
   @VisibleForTesting
-  public ConcurrentMap<String, AtomicReference<ManagedChannel>> getPerHostChannelMap() {
+  public ConcurrentMap<InstanceDetails.Id, AtomicReference<ManagedChannel>> getPerHostChannelMap() {
     return perHostChannelMap;
   }
 
   @VisibleForTesting
-  public ConcurrentMap<String, AtomicReference<InterNodeRpcServiceStub>> getPerHostClientStubMap() {
+  public ConcurrentMap<InstanceDetails.Id, AtomicReference<InterNodeRpcServiceStub>> getPerHostClientStubMap() {
     return perHostClientStubMap;
   }
 
@@ -113,9 +112,8 @@ public class GRPCConnectionManager {
    * @return The stub object.
    */
   public InterNodeRpcServiceStub getClientStubForHost(
-      final String remoteHost) {
-    final AtomicReference<InterNodeRpcServiceStub> stubAtomicReference =
-        perHostClientStubMap.get(remoteHost);
+      final InstanceDetails remoteHost) {
+    final AtomicReference<InterNodeRpcServiceStub> stubAtomicReference = perHostClientStubMap.get(remoteHost);
     if (stubAtomicReference != null) {
       return stubAtomicReference.get();
     }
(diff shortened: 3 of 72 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/DeciderActionPriorityReader.java
```diff
@@ -45,8 +45,7 @@ public static void updateDeciderActionPriorityOrder() {
         try {
             deciderActionPriorityOrder = mapper.readValue(new File(configFilePath), HashMap.class);
         } catch (Exception e) {
-            LOG.error("Could not read the Decider Config File");
-            e.printStackTrace();
+            LOG.error("Could not read the Decider Config File: " + configFilePath);
```
sidheart, line 48: Log.error("Could not read the Decider Config File: {}", configFilePath, e)

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decision.java
```diff
@@ -19,6 +19,8 @@
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.FlowUnitMessage;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.core.GenericFlowUnit;
 
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.util.InstanceDetails;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.reader.ClusterDetailsEventProcessor;
```
sidheart, line 23: This import line was added, but I don't see CDEP being added below

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
```diff
@@ -159,28 +155,36 @@ private ManagedChannel getChannelForHost(final String remoteHost) {
...
     return shouldUseHttps ? buildSecureChannel(remoteHost) : buildInsecureChannel(remoteHost);
   }
 
-  private ManagedChannel buildInsecureChannel(final String remoteHost) {
-    return ManagedChannelBuilder.forAddress(remoteHost, this.port).usePlaintext().build();
+  private ManagedChannel buildInsecureChannel(final InstanceDetails remoteHost) {
+    int port = this.port != 0 ? this.port : remoteHost.getGrpcPort();
+    if (port == -1) {
+      throw new IllegalArgumentException("Invalid port for grpc: " + port);
+    }
```
sidheart, line 173: nit: factor this out into a function since it's used in multiple places

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
```diff
@@ -54,17 +55,15 @@
   private File trustedCasFile;
 
   /**
-   * Map of remote host to a Netty channel to that host.
+   * Map of remote hostId to a Netty channel to that host.
    */
-  private ConcurrentMap<String, AtomicReference<ManagedChannel>> perHostChannelMap =
-      new ConcurrentHashMap<>();
+  private ConcurrentMap<InstanceDetails.Id, AtomicReference<ManagedChannel>> perHostChannelMap = new ConcurrentHashMap<>();
```
ktkrg, line 60: I agree with most of the refactor points mentioned in the description of the PR, however, can you explain why we need to use node ids instead of ip+port combo? It would've achieved the same goal, unless I'm missing something.

yojs: You are right, ip+port will have the same affect. I used the node-id as it is a parameter, on its own (without combining with anything else), is guaranteed to be unique in a cluster.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/DeciderActionPriorityReader.java
yojs: This file has been removed in the latest update to master.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decision.java
yojs: good catch !

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
ktkrg: Thanks for answering my question. I'm just thinking about the usecases where ip+port wouldn't change but node id would and if such scenarios are handled in the code if they were not handled in the past. Will take a look again.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ResourceHeatMapGraphTest.java
```diff
@@ -603,10 +603,13 @@ private void verifyFullNodeTemperatureProfile(String resp) {
...
-        .getAsJsonArray(ALL_TEMPERATURE_DIMENSIONS)
-        .get(0)
-        .getAsJsonObject()
-        .getAsJsonArray(NodeLevelDimensionalSummary.SUMMARY_TABLE_NAME);
+        .getAsJsonArray(ALL_TEMPERATURE_DIMENSIONS);
+    // System.out.println("ALL_TEMPERATURE_DIMENSIONS" + json);
+
+    json.get(0)
+    .getAsJsonObject()
+    .getAsJsonArray(NodeLevelDimensionalSummary.SUMMARY_TABLE_NAME);
```
ktkrg, line 611: return value is being ignored here. Do we need this change?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
```diff
@@ -113,6 +114,11 @@ public void overrideProperty(String key, String value) {
     settings.setProperty(key, value);
   }
 
+  @VisibleForTesting
+  public @Nullable String getProperty(String key) {
```
sidheart, line 118: This annotation should go above the method if you're saying that the returned String may be null

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/InstanceDetails.java
```diff
@@ -15,48 +15,166 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hashCode(ip);
+    }
+
+    @Override
+    public String toString() {
+      return "" + ip;
```
sidheart, line 57: I'm curious, what does this do differently than just directly returning ip?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/messages/DataMsg.java
```diff
@@ -19,23 +19,23 @@
...
-  List<String> destinationNodes;
+  String sourceGraphNode;
+  List<String> destinationGraphNodes;
   List<? extends GenericFlowUnit> flowUnits;
 
   public DataMsg(
-      String sourceNode, List<String> destinationNode, List<? extends GenericFlowUnit> flowUnits) {
-    this.sourceNode = sourceNode;
-    this.destinationNodes = destinationNode;
+          String sourceGraphNode, List<String> destinationNode, List<? extends GenericFlowUnit> flowUnits) {
```
sidheart, line 27: nit: destinationGraphNodes?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
```diff
@@ -73,7 +72,7 @@
 
   public GRPCConnectionManager(final boolean shouldUseHttps) {
     this.shouldUseHttps = shouldUseHttps;
-    this.port = Util.RPC_PORT;
+    this.port = -1;
```
sidheart, line 75: Is there a reason you initialize to -1 instead of 0? 0 would make the default behavior the same as before this commit right?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/net/SubscriptionManagerTest.java
```diff
@@ -2,16 +2,19 @@
...
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.mockito.MockitoAnnotations;
 
-
+// TODO: remove
```
sidheart, line 16: It's ok to remove this @Ignore and change grpcConnectionManager = new GRPCConnectionManager(false). TLS semantics are tested elsewhere.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/RcaEnabledSamplerTest.java
```diff
@@ -39,14 +39,14 @@ public void testIsRcaEnabled() {
...
-                ClusterDetailsEventProcessorTestHelper.newNodeDetails("", "", false);
+                ClusterDetailsEventProcessorTestHelper.newNodeDetails("nodex", "127.0.0.1", false);
 
         ClusterDetailsEventProcessor clusterDetailsEventProcessor = new ClusterDetailsEventProcessor();
         clusterDetailsEventProcessor.setNodesDetails(Collections.singletonList(details));
         appContext.setClusterDetailsEventProcessor(clusterDetailsEventProcessor);
 
         assertFalse(uut.isRcaEnabled());
-        details = ClusterDetailsEventProcessorTestHelper.newNodeDetails("", "", true);
+        details = ClusterDetailsEventProcessorTestHelper.newNodeDetails("nodey", "127.0.0.2", true);
```
sidheart, line 49: It's like a pet name for the node! :)

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/store/rca/cluster/BaseClusterRcaTest.java
```diff
@@ -310,7 +310,10 @@ private boolean compareResourceSummary(Resource resource, HotResourceSummary res
   }
 
   private boolean compareNodeSummary(String nodeId, Resource resource, HotNodeSummary nodeSummary) {
-    if (!nodeId.equals(nodeSummary.getNodeID()) || nodeSummary.getHotResourceSummaryList().isEmpty()) {
+    if (!nodeId.equals(nodeSummary.getNodeID().toString())) {
+      return false;
```
sidheart, line 314: nit: why split this if we aren't distinguishing between the two cases?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
yojs: done !

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
yojs: makes sense. Please take a look.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ResourceHeatMapGraphTest.java
yojs: ya. I was running tests and left it like that. good point !

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/InstanceDetails.java
yojs: you mean why we don't have `getIp()` instead ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/messages/DataMsg.java
yojs: 👍

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
yojs: You are right. That was a miss on my part.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/net/SubscriptionManagerTest.java
yojs: I added the change for it to get the right files.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/RcaEnabledSamplerTest.java
yojs: :)

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/store/rca/cluster/BaseClusterRcaTest.java
yojs: It was just that I was debugging and it makes it easier to put breakpoints to know which branch is returning the error. :)

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/InstanceDetails.java
sidheart: I mean why `return "" + ip` instead of just `return ip`?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
sidheart: I think you missed this one, if this is changed I can +1 the diff

yojs: I missed this one. Pushed the changes again.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/util/InstanceDetails.java
yojs: Yes you are right. It's a no-op for a string

Issue #313, opened 2020-07-28
Consume current queue capacity from NodeConfigCollector


aditjind:
Consume current queue capacity from NodeConfigCollector.

A node level cache is already implemented which has updated values of the queues and caches. We need to read these values in the Decider implementation for the decider to compute the updated value of the Queue.

PR #314, opened 2020-07-28
Consume Queue Capacity from Node Config Cache.

aditjind:
*Issue #:* #313 313

*Description of changes:* Return the updated value in the cache. In case the value is not updated in the cache, don't take any action since we don't want to update the value as we don't have the state and we can wait for the cache to have an updated value. 

*Tests:* Manual testing. 

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
@@ -93,11 +93,11 @@ public class QueueHealthDecider extends Decider {
    * <p>Action relevance decided based on user configured priorities for now, this can be modified
    * to consume better signals going forward.
    */
-  private Action computeBestAction(NodeKey esNode, ResourceEnum threadpool) {
+  private Action computeBestAction(NodeKey esNode, ResourceEnum threadPool) {
     Action action = null;
     int currQueueCapacity;
     try {
-        currQueueCapacity = getNodeQueueCapacity(esNode, threadpool);
+        currQueueCapacity = getNodeQueueCapacity(esNode, threadPool);
     } catch (Exception e) {
       // No action if value not present in the cache.
       // Assumption here is the cache has been wiped off due to
@@ -108,7 +108,7 @@ public class QueueHealthDecider extends Decider {
 
     for (String actionName : actionsByUserPriority) {
       action =
-        getAction(actionName, esNode, threadpool, currQueueCapacity, true);
+        getAction(actionName, esNode, threadPool, currQueueCapacity, true);
       if (action != null) {
         break;
       }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
@@ -95,9 +95,20 @@ public class QueueHealthDecider extends Decider {
    */
   private Action computeBestAction(NodeKey esNode, ResourceEnum threadpool) {
     Action action = null;
+    int currQueueCapacity;
+    try {
+        currQueueCapacity = getNodeQueueCapacity(esNode, threadpool);
+    } catch (Exception e) {
+      // No action if value not present in the cache.
+      // Assumption here is the cache has been wiped off due to
+      // unforeseen events and we dont want to trigger any action.
+      LOG.error("Exception while reading values from Node Config Cache", e);
+      return null;
+    }
+
     for (String actionName : actionsByUserPriority) {
       action =
-        getAction(actionName, esNode, threadpool, getNodeQueueCapacity(esNode, threadpool), true);
+        getAction(actionName, esNode, threadpool, currQueueCapacity, true);
       if (action != null) {
         break;
       }
@@ -106,9 +117,6 @@ public class QueueHealthDecider extends Decider {
   }
 
   private Action getAction(String actionName, NodeKey esNode, ResourceEnum threadPool, int currCapacity, boolean increase) {
-    if (currCapacity == -1) {
-      return null;
-    }
     switch (actionName) {
       case ModifyQueueCapacityAction.NAME:
         return configureQueueCapacity(esNode, threadPool, currCapacity, increase);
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -91,11 +92,11 @@ private void configureActionPriority() {
    * <p>Action relevance decided based on user configured priorities for now, this can be modified
    * to consume better signals going forward.
    */
-  private Action computeBestAction(NodeKey esNode, ResourceEnum threadPool) {
+  private Action computeBestAction(NodeKey esNode, Resource resource) {
     Action action = null;
     for (String actionName : actionsByUserPriority) {
       action =
-          getAction(actionName, esNode, threadPool, getNodeQueueCapacity(esNode, threadPool), true);
+        getAction(actionName, esNode, resource.getResourceEnum(), getNodeQueueCapacity(esNode, resource), true);
```
rguo-aws, line 99: resource object read from QueueRejctionRca is threadpool + queue rejection. That is not the metric we want to read here. What we want to read from node config cache is threadpool + queue capacity

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -120,11 +124,15 @@ private ModifyQueueCapacityAction configureQueueCapacity(NodeKey esNode, Resourc
     return null;
   }
 
-  private int getNodeQueueCapacity(NodeKey esNode, ResourceEnum threadPool) {
-    // TODO: use NodeConfigurationRca to return capacity, for now returning defaults
-    if (threadPool.equals(ResourceEnum.SEARCH_THREADPOOL)) {
-      return 1000;
+  private int getNodeQueueCapacity(NodeKey esNode, Resource resource) {
```
rguo-aws, line 127: as I mentioned above, we don't have to change the input argument of this function. What we can do instead is to map queue rejection -> queue capacity

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -17,31 +17,32 @@
...
 public class QueueHealthDecider extends Decider {
 
+  private static final Logger LOG = LogManager.getLogger(Decider.class);
   public static final String NAME = "queue_health";
 
   private QueueRejectionClusterRca queueRejectionRca;
   List<String> actionsByUserPriority = new ArrayList<>();
   private int counter = 0;
 
   public QueueHealthDecider(long evalIntervalSeconds, int decisionFrequency, QueueRejectionClusterRca queueRejectionClusterRca) {
```
rguo-aws, line 45: remove QueueRejectionClusterRca from the constructor as we don't consume anything directly from QueueRejectionClusterRca

vigyasharma: Don't we rely on `QueueRejectionClusterRca` flow units to check whether a thread pool queue is healthy/unhealthy - @rguo-aws ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -121,10 +126,16 @@ private ModifyQueueCapacityAction configureQueueCapacity(NodeKey esNode, Resourc
...
+              .setResourceEnum(threadPool)
+              .setMetricEnum(MetricEnum.QUEUE_CAPACITY).build());
+    } catch (Exception e) {
+      LOG.error("Exception while reading values from Node Config Cache", e);
     }
-    return 100;
+    // No action if value not present in the cache.
+    // Assumption here is the cache has been wiped off due to
+    // unforeseen events and we dont want to trigger any action.
+    return -1;
```
vigyasharma, line 139: Minor: This is bit of a code smell. Can we raise an exception here instead and handle it gracefully at the caller? Instead of overloading the meaning of return value?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
aditjind: Yes resolved this with ruizhen!

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
aditjind: Done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -121,10 +134,8 @@ private ModifyQueueCapacityAction configureQueueCapacity(NodeKey esNode, Resourc
   }
 
   private int getNodeQueueCapacity(NodeKey esNode, ResourceEnum threadPool) {
-    // TODO: use NodeConfigurationRca to return capacity, for now returning defaults
-    if (threadPool.equals(ResourceEnum.SEARCH_THREADPOOL)) {
-      return 1000;
-    }
-    return 100;
+      return (int) getAppContext().getNodeConfigCache().get(esNode, Resource.newBuilder()
```
khushbr, line 137: This will return an `IllegalArgumentException` if the NodeKey is not present. We should add handling for it.

aditjind: I have handled exceptions in the caller of this function. It will result in no action being taken on the cluster.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -121,10 +134,8 @@ private ModifyQueueCapacityAction configureQueueCapacity(NodeKey esNode, Resourc
   }
 
   private int getNodeQueueCapacity(NodeKey esNode, ResourceEnum threadPool) {
```
rguo-aws, line 136: nit: should we add a throws keyword here ?

PR #315, opened 2020-07-28
Batch Metrics API

ricardolstephen:
*Issue #:*
#335 

*Description of changes:*
Batch Metrics API implementation
Associated PA change: https://github.com/opendistro-for-elasticsearch/performance-analyzer/pull/159

*Tests:*
* Test config and cluster config requests
* Test no batch metrics files when disabled
* Test 85 metrics files when batch metrics enabled
* Test querying 5min of data
* Test querying into the future fails
* Test querying past the retention period fails
* Test with sampling periods of 5s, 10s, 20s
* Test that metricsdb files older than the retention period are cleaned up
* Test that files are cleaned up when batch metrics is disabled
* Test another enable-disable cycle
* Test that the max datapoints value (100800) is respected
* Test that invalid "batch-metrics-retention-period-minutes" values (0, 61) are rejected and valid values (1, 60) are accepted.
* Test that metricsdb files are not cleaned up if "cleanup-metrics-db-files" is disabled (verify this across multiple enable-disable cycles)
* Test that querying past the retention period fails even when "cleanup-metrics-db-files" is disabled
* Test that nodes with and without the batch metrics feature can co-exist in a cluster:
  * Test that nodes with batch metrics can join the cluster and correctly enable and disable it from the cluster state.
  * Test that both types of nodes can correctly enable and disable rca and logging from the cluster state without affecting the state of batch metrics
* Test that missing or corrupt metricsdb files result in the appropriate error messages.


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
```diff
@@ -66,6 +67,7 @@
 
   private static final int EXCEPTION_QUEUE_LENGTH = 1;
   public static final String QUERY_URL = "/_opendistro/_performanceanalyzer/metrics";
+  public static final String BATCH_METRICS_URL = "/_opendistro/_performanceanalyzer/batch";
```
yojs, line 70: Please change the README accordingly to document the new url and details for this

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
```diff
@@ -44,6 +45,9 @@
   private static final int DELETION_INTERVAL_MAX = 60;
   private static final String HTTPS_ENABLED = "https-enabled";
   private static final String WRITER_QUEUE_SIZE = "writer-queue-size";
+  private static final String BATCH_METRICS_RETENTION_PERIOD = "batch-metrics-retention-period";
+  private static final long BATCH_METRICS_RETENTION_PERIOD_DEFAULT = 7;  // 7 minutes of metrics
+  private static final long BATCH_METRICS_RETENTION_PERIOD_LIMIT = 60;
```
yojs, line 50: Can we add the time unit in the variable name such as
`BATCH_METRICS_RETENTION_PERIOD_DEFAULT_MINS` and so on ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
```diff
@@ -44,6 +45,9 @@
   private static final int DELETION_INTERVAL_MAX = 60;
   private static final String HTTPS_ENABLED = "https-enabled";
   private static final String WRITER_QUEUE_SIZE = "writer-queue-size";
+  private static final String BATCH_METRICS_RETENTION_PERIOD = "batch-metrics-retention-period";
+  private static final long BATCH_METRICS_RETENTION_PERIOD_DEFAULT = 7;  // 7 minutes of metrics
```
yojs, line 49: Also can we add javadoc comments describing the variables ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
```diff
@@ -59,6 +63,8 @@
   private Properties settings;
   private final String configFilePath;
 
+  private long batchMetricsRetentionPeriod;
```
yojs, line 66: Please add the timeunit as the variable name

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
```diff
@@ -156,11 +178,12 @@ private PluginSettings(String cfPath) {
     }
     LOG.info(
         "Config: metricsLocation: {}, metricsDeletionInterval: {}, httpsEnabled: {},"
-            + " cleanup-metrics-db-files: {}",
+            + " cleanup-metrics-db-files: {}, batch-metrics-retention-period: {}",
```
yojs, line 181: Let's add the unit here as well, so that it is reflected in the logs

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
```diff
@@ -271,4 +294,26 @@ private void loadMetricsDBFilesCleanupEnabled() {
...
+  private void loadBatchMetricsRetentionPeriodFromConfig() {
+    if (!settings.containsKey(BATCH_METRICS_RETENTION_PERIOD)) {
+      return;
+    }
+
+    long parsedRetentionPeriod;
+    try {
+      parsedRetentionPeriod = Long.parseLong(settings.getProperty(BATCH_METRICS_RETENTION_PERIOD));
+      if (parsedRetentionPeriod <= 0 || parsedRetentionPeriod > BATCH_METRICS_RETENTION_PERIOD_LIMIT) {
+        throw new InvalidParameterException();
```
yojs, line 307: We are throwing an exception here but maybe we should just log an error and use default values ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
```diff
@@ -271,4 +294,26 @@ private void loadMetricsDBFilesCleanupEnabled() {
...
+      }
+      batchMetricsRetentionPeriod = parsedRetentionPeriod;
+    } catch (NumberFormatException | InvalidParameterException e) {
+      LOG.error(
+              (Supplier<?>)
+                      () ->
+                              new ParameterizedMessage(
+                                      "Invalid batch-metrics-retention-period. Using default value {}.", batchMetricsRetentionPeriod),
+              e);
```
yojs, line 316: same here. Log error and then we should just use the defaults.

Also you can log error simply as:

`LOG.error("my error message: {}", myParam)`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -84,6 +85,15 @@ public MetricsDB(long windowStartTime) throws Exception {
     create = DSL.using(conn, SQLDialect.SQLITE);
   }
 
+  public static MetricsDB fetchExisting(long windowStartTime) throws Exception {
+    String filePath = PluginSettings.instance()
+            .getSettingValue(DB_FILE_PREFIX_PATH_CONF_NAME, DB_FILE_PREFIX_PATH_DEFAULT) + windowStartTime;
+    if (!(new File(filePath)).exists()) {
+      throw new FileNotFoundException(String.format("MetricsDB file %s could not be found.", filePath));
```
yojs, line 92: The new changes should not crash the system. we should handle this gracefully.

Also, can we add javadoc comments to public methods ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -239,6 +249,21 @@ public void putMetric(Metric<Double> metric, Dimensions dimensions, long windowS
     return create.select().from(DSL.table(metric)).fetch();
   }
 
+  public Result<Record> queryMetric(String metric, Collection<String> dimensions, int limit) {
+    if (!DBUtils.checkIfTableExists(create, metric)) {
+      return null;
```
yojs, line 254: Can we return empty Object instead of null ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -239,6 +249,21 @@ public void putMetric(Metric<Double> metric, Dimensions dimensions, long windowS
...
+    }
+    if (limit < 0) {
+      throw new IllegalArgumentException("Limit must be non-negative");
```
yojs, line 257: This is too late to validate if the limit is negative. Maybe we should do it up in the call handling stack ?

ricardolstephen: Such a check is enforced multiple times in QueryBatchRequestHandler. The check is also included here in MetricsDB since adding this function expands the API of MetricsDB, and such a check is required for the new expanded API.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
ricardolstephen: We return a null here to stay consistent with the other queryMetric method:
public Result<Record> queryMetric(List<String> metrics, List<String> aggregations, List<String> dimensions) throws Exception

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -201,8 +219,41 @@ public void trimOldSnapshots() throws Exception {
     trimMap(shardRqMetricsMap, RQ_SNAPSHOTS);
     trimMap(httpRqMetricsMap, HTTP_RQ_SNAPSHOTS);
     trimMap(masterEventMetricsMap, MASTER_EVENT_SNAPSHOTS);
-    trimDatabases(
-        metricsDBMap, MAX_DATABASES, PluginSettings.instance().shouldCleanupMetricsDBFiles());
+
+    boolean deleteDBFiles = PluginSettings.instance().shouldCleanupMetricsDBFiles();
+    if (metricsDBMap.size() > MAX_DATABASES) {
+      Map.Entry<Long, MetricsDB> lowestEntry = metricsDBMap.pollFirstEntry();
```
sidheart, line 225: by lowest do you mean oldest?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -201,8 +219,41 @@ public void trimOldSnapshots() throws Exception {
...
+      if (lowestEntry != null) {
+        Long key = lowestEntry.getKey();
+        MetricsDB value = lowestEntry.getValue();
+        value.remove();
+        if (deleteDBFiles && !batchMetricsDBSet.contains(key)) {
+          value.deleteOnDiskFile();
```
sidheart, line 231: This logic should be placed into the trimDatabases function

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -696,6 +750,24 @@ public void deleteDBs() throws Exception {
...
+   * This is called by operations outside of the ReaderMetricsProcessor.
+   *
+   * @return A list of the timestamps associated with MetricsDB files. The oldest of the MetricsDB files typically
+   *     have a lifetime of ~SAMPLING_INTERVAL seconds (no less than SAMPLING_INTERVAL/2 seconds). Null if batch
+   *     metrics is disabled.
+   */
+  public NavigableSet<Long> getBatchMetrics() {
+    if (batchMetricsEnabled) {
+      TreeSet<Long> batchMetricsDBSetCopy = new TreeSet<>(batchMetricsDBSet.clone());
+      while (batchMetricsDBSetCopy.size() > PluginSettings.instance().getBatchMetricsRetentionPeriod() * 12) {
```
sidheart, line 763: Add a comment about why you're using 12 b/c of the 5s snapshots. Also make it evident that the retention period is in minutes

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -201,8 +219,41 @@ public void trimOldSnapshots() throws Exception {
...
+      batchMetricsDBSet.clear();
+    }
+    readBatchMetricsEnabledFromConf();
+    // The (retentionPeriod * 12 + 1)'th database can be safely removed, since getBatchMetrics never returns more than
+    // the (retentionPeriod * 12) freshest metrics files.
+    if (batchMetricsDBSet.size() > PluginSettings.instance().getBatchMetricsRetentionPeriod() * 12 + 1) {
+      Long timestamp = batchMetricsDBSet.pollFirst();
+      if (deleteDBFiles && !metricsDBMap.containsKey(timestamp)) {
+        MetricsDB.deleteOnDiskFile(timestamp);
+      }
```
sidheart, line 255: This should be a while loop

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -201,8 +219,41 @@ public void trimOldSnapshots() throws Exception {
...
+    if (metricsDBMap.size() > MAX_DATABASES) {
+      Map.Entry<Long, MetricsDB> lowestEntry = metricsDBMap.pollFirstEntry();
+      if (lowestEntry != null) {
+        Long key = lowestEntry.getKey();
+        MetricsDB value = lowestEntry.getValue();
+        value.remove();
+        if (deleteDBFiles && !batchMetricsDBSet.contains(key)) {
+          value.deleteOnDiskFile();
+        }
+      }
```
sidheart, line 233: This should be a while loop

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
ricardolstephen: Will add this documentation to the batchMetricsRetentionPeriod variable (this is what was done for the other parameters also).

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
ricardolstephen: Yes, that would be a good idea.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
ricardolstephen: Yes

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
ricardolstephen: Yes

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -84,6 +85,15 @@ public MetricsDB(long windowStartTime) throws Exception {
     create = DSL.using(conn, SQLDialect.SQLITE);
   }
 
+  public static MetricsDB fetchExisting(long windowStartTime) throws Exception {
+    String filePath = PluginSettings.instance()
```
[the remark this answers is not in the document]
ricardolstephen: getDBFilePath can't be called from a static context -- it's an instance method that depends on the windowStartTime field.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
ricardolstephen: This change would not crash the system. The QueryBatchRequestHandler (the only place where this is used) catches it exception and handles it appropriately. Also, this code here isn't introducing any new error cases -- it has the same abstraction of throwing an error if there is any issue accessing the file or creating a jdbc connection to it as the MetricsDB constructor.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -259,6 +284,19 @@ public void deleteOnDiskFile() {
     }
   }
 
+  public static void deleteOnDiskFile(long windowStartTime) {
+    String dbFilePath = PluginSettings.instance()
+            .getSettingValue(DB_FILE_PREFIX_PATH_CONF_NAME, DB_FILE_PREFIX_PATH_DEFAULT) + windowStartTime;
```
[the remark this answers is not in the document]
ricardolstephen: getDBFilePath can't be called from a static context -- it's an instance method that depends on the windowStartTime field.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: Yes, will change the variable names to reflect this.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: trimDatabases currently doesn't accept a parameter that would allow us to retain the db if it is in batchMetricsDBSet. trimDatabases could be modified to include such logic. However, such a change wouldn't really contribute much, as trimDatabases is only used in timOldSnapshots. Moving it out of trimOldSnapshots would also not even help with testing, as there are two other different trimming-related procedures in trimOldSnapshots.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: Agreed

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: Agreed

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: An understanding of the 5s snapshots should be obvious to the reader of this code. The connection to the 12 should be pretty readily understood.
Agreed on the minutes unit for the retention period -- the method itself will be renamed to say "Minutes". This should further make the connection for the first part obvious.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
ricardolstephen: Agreed about the javadocs.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
ricardolstephen: Joydeep, if any exception is thrown here, the batchMetricsRetentionPeriodMinutes value will not have been modified (since it is in the last line of the try block), and so the default value will be retained. Additionally, the method itself catches the thrown Exception and logs it.

ricardolstephen: Sid, Sruti's pr is only about disabling the reader if the config file cannot be found. It was not about disabling the reader if values cannot be correctly parsed from the config file -- using the default value for these cases is the desired behavior.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
ricardolstephen: Agreed, documentation has been added to the README in the performance-analyzer repository.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
ricardolstephen: Agreed, updated the error logging call.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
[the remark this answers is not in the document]
sidheart: The way you have it now, the same logic is in 2 different places. Change getDBFilePath to call a function like

static String getActualFilePath(long windowStartTime) {
 // PluginSettings.instance().getSettingValue(...)
}

have getDBFilePath and fetchExisting call this method

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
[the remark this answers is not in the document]
sidheart: See above

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
```diff
@@ -0,0 +1,309 @@
...
+        for (metricIndex += 1; metricIndex < numMetrics; metricIndex++) {
+          metric = metrics.get(metricIndex);
+          results = db.queryMetric(metric, MetricsModel.ALL_METRICS.get(metric).dimensionNames, maxDatapoints);
+          if (results != null) {
+            maxDatapoints -= results.size();
+            if (maxDatapoints == 0) {
+              throw new InvalidParameterException(String.format("requested data exceeds the %d datapoints limit", DEFAULT_MAX_DATAPOINTS));
+            }
+            builder.append(",\"");
+            builder.append(metric);
```
sidheart, line 263: I've already commented on this to Ricardo. I think it's better to avoid tight coupling between the JSON response construction and the actual sampling and querying of the metrics. 

However, since today there are no components that would actually need the batching logic outside of the REST layer, and since this is an L0 solution, I've said that I'd note my concern here and +1.

sidheart: Does anyone else have a strong opinion on this? @yojs @vigyasharma

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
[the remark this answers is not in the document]
ricardolstephen: Sounds reasonable

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
```diff
@@ -0,0 +1,309 @@
...
+    Long metricsTimestamp = batchMetrics.ceiling(startTime);
+    if (metricsTimestamp != null && metricsTimestamp < endTime) {
+      maxDatapoints = appendMetrics(metricsTimestamp, metrics, responseJson, maxDatapoints);
+      metricsTimestamp = metricsTimestamp - metricsTimestamp % samplingPeriod + samplingPeriod;
+      metricsTimestamp = batchMetrics.ceiling(metricsTimestamp);
+      while (metricsTimestamp != null && metricsTimestamp < endTime) {
+        responseJson.append(",");
+        maxDatapoints = appendMetrics(metricsTimestamp, metrics, responseJson, maxDatapoints);
+        metricsTimestamp = metricsTimestamp - metricsTimestamp % samplingPeriod + samplingPeriod;
+        metricsTimestamp = batchMetrics.ceiling(metricsTimestamp);
```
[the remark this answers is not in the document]
ricardolstephen: This is round down.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
ricardolstephen: If we find a use for such decoupling, perhaps we can modify the implementation with such changes. However, for now, this approach will work, and it removes the need for additional investigation (for example, this would increase processing time, and stringifying a Batch Metrics Response object would double the peak memory usage of this component).

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
```diff
@@ -0,0 +1,309 @@
...
+
+  private static final Logger LOG = LogManager.getLogger(QueryBatchRequestHandler.class);
+
+  private static final int TIME_OUT_VALUE = 2;
+  private static final TimeUnit TIME_OUT_UNIT = TimeUnit.SECONDS;
+  private NetClient netClient;
+  MetricsRestUtil metricsRestUtil;
+
+  public static final int DEFAULT_MAX_DATAPOINTS = 100800;  // Must be non-negative
+  public static final long DEFAULT_SAMPLING_PERIOD = 5000;  // Must be a multiple of 5000
```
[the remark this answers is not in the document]
ricardolstephen: Sure

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
```diff
@@ -0,0 +1,309 @@
...
+    builder.append("\":{");
+    MetricsDB db = MetricsDB.fetchExisting(timestamp);
+    for (int metricIndex = 0, numMetrics = metrics.size(); metricIndex < numMetrics; metricIndex++) {
+      String metric = metrics.get(metricIndex);
+      Result<Record> results = db.queryMetric(metric, MetricsModel.ALL_METRICS.get(metric).dimensionNames, maxDatapoints);
+      if (results != null) {
+        maxDatapoints -= results.size();
+        if (maxDatapoints == 0) {
+          throw new InvalidParameterException(String.format("requested data exceeds the %d datapoints limit", DEFAULT_MAX_DATAPOINTS));
+        }
```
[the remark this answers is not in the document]
ricardolstephen: `results.size() cannot be ` cannot be > `maxDatapoints`, since `maxDatapoints` is supplied as the limit when invoking `db.queryMetric`.

`DEFAULT_MAX_DATAPOINTS` cannot be overridden. It is an internal safeguard to ensure that the user does not break their system by exceeding a memory limit. Even if we allowed the user to configure the max datapoints, we would still need an internal limit to ensure no excessive values are supplied. Until more clear internal limits are identified, we ought to leave configurability off the table in favor of safety. Additionally, these changes may further be obviated if pagination is introduced in the future.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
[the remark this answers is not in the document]
sidheart: If metricsTimestamp is 1 with a samplingPeriod of 5 then the new metricsTimestamp becomes 5 and you return all metrics with a timestamp bigger than 5. How is that rounding down? Is this what you intended?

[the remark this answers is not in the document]
ricardolstephen: The starttime and enddtime's supplied by the user are rounded down in line 193-194. The code here is for when you're calculating all the timestamps you want to collect data from.

This is what the code here does:
1. metricsTimestamp = 0, samplingperiod = 30
2. Line 292 -> metricsTimestamp = 30
3. Line 293 -> metricsTimestamp = 65 (imagine the data between 30-60 is missing).
4. Next cycle
5. Line 292 -> metricsTimestamp = 90
...

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandlerTest.java
```diff
@@ -0,0 +1,516 @@
...
+import java.util.TreeSet;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.mockito.ArgumentMatchers;
+import org.mockito.Mockito;
+
+public class QueryBatchRequestHandlerTest {
```
yojs, line 53: can we please make all method names camelCase ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -84,6 +89,21 @@ public MetricsDB(long windowStartTime) throws Exception {
     create = DSL.using(conn, SQLDialect.SQLITE);
   }
 
+  /**
+   * Returns a MetricsDB handle associated with an existing metricsdb file.
```
yojs, line 93: Will calling it the latest metricDB file be more appropriate ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -201,8 +219,43 @@ public void trimOldSnapshots() throws Exception {
     trimMap(shardRqMetricsMap, RQ_SNAPSHOTS);
     trimMap(httpRqMetricsMap, HTTP_RQ_SNAPSHOTS);
     trimMap(masterEventMetricsMap, MASTER_EVENT_SNAPSHOTS);
-    trimDatabases(
-        metricsDBMap, MAX_DATABASES, PluginSettings.instance().shouldCleanupMetricsDBFiles());
+
```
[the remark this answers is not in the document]
yojs: The better way might be to move lines 223 to 258 in the trimDatabase method. what do you think ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -84,6 +85,15 @@ public MetricsDB(long windowStartTime) throws Exception {
     create = DSL.using(conn, SQLDialect.SQLITE);
   }
 
+  public static MetricsDB fetchExisting(long windowStartTime) throws Exception {
```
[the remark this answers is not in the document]
ricardolstephen: We only marked those other callers as "VisibleForTesting" because we expanded their visibility from private to public simply for the sake of testing. However, even if they were private, this method would have to have public visibility for them to be able to access it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
```diff
@@ -271,4 +295,21 @@ private void loadMetricsDBFilesCleanupEnabled() {
...
+
+  private void loadBatchMetricsRetentionPeriodMinutesFromConfig() {
+    if (!settings.containsKey(BATCH_METRICS_RETENTION_PERIOD_MINUTES)) {
+      return;
+    }
+
+    try {
+      long parsedRetentionPeriod = Long.parseLong(settings.getProperty(BATCH_METRICS_RETENTION_PERIOD_MINUTES));
+      if (parsedRetentionPeriod <= 0 || parsedRetentionPeriod > BATCH_METRICS_RETENTION_PERIOD_MINUTES_LIMIT) {
+        throw new InvalidParameterException();
```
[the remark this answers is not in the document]
ricardolstephen: That is what this code does. batchMetricsRetentionPeriodMinutes was assigned a default value in the ctor. The code here only updates that value if it was able to properly parse a new valid value from the properties file.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandlerTest.java
ricardolstephen: The naming here makes use of the conventions outlined here:
https://google.github.io/styleguide/javaguide.html#s5.2.3-method-names
There is also precedence here:
https://github.com/opendistro-for-elasticsearch/performance-analyzer/blob/master/src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/http_action/config/PerformanceAnalyzerResourceProviderTest.java

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -201,8 +219,43 @@ public void trimOldSnapshots() throws Exception {
...
+      }
+    }
+    // Flush any tracking batch metrics if batch metrics is disabled. Note, in order to ensure that batch metrics
+    // consumers have had at least one cycle to use any metrics they may be holding, this flush is done before
+    // re-reading the config file to update the state of the batch metrics feature.
+    if (!batchMetricsEnabled && !batchMetricsDBSet.isEmpty()) {
+      if (deleteDBFiles) {
+        for (Long timestamp : batchMetricsDBSet) {
+          if (!metricsDBMap.containsKey(timestamp)) {
+            MetricsDB.deleteOnDiskFile(timestamp);
```
yojs, line 242: Above we have gone through all the metricsDBMap and would have deleted all files but the two most recent ones. Can you add code comment explaining why are we doing this again ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -201,8 +219,43 @@ public void trimOldSnapshots() throws Exception {
...
+          }
+        }
+      }
+      batchMetricsDBSet.clear();
+    }
+    readBatchMetricsEnabledFromConf();
```
yojs, line 248: This sounds like this call may not belong here or the name of the enclosing method `trimOldSnapshots` might be changed to something more relevant indicating that the config is re-read here ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -289,6 +342,9 @@ private void emitMetrics(long currWindowStartTime) throws Exception {
 
     metricsDB.commit();
     metricsDBMap.put(prevWindowStartTime, metricsDB);
+    if (batchMetricsEnabled) {
+      batchMetricsDBSet.add(prevWindowStartTime);
```
yojs, line 346: SO we store the snapshot everytime we emit metrics, at query time we clone the snapshot (so we double the snapshot) and then we trim it. This is very memory intensive. We might just want to read the files directly from disk when the API is called ?

ricardolstephen: We do not clone the snapshot at query time. We read the file itself. We use the batchMetricsDBSet to keep track of the timestamps of all the available dbs we have on disk, and then when we want to make use of these files during query handling, we call MetricsDB.fetchExisting, which simply creates a new connection to that on-disk database.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
ricardolstephen: It would not. There may be 30 metricsdb files on disk. The user would call this method with the timestamp associated with one of them, and this method would return a MetricsDB handle to that file.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -84,6 +89,21 @@ public MetricsDB(long windowStartTime) throws Exception {
...
   }
 
+  /**
+   * Returns a MetricsDB handle associated with an existing metricsdb file.
+   *
+   * @param windowStartTime the timestamp associated with an existing metricsdb file
+   * @return a MetricsDB handle associated with the metricsdb file
+   * @throws Exception if the metricsdb file does not exist or is invalid
+   */
+  public static MetricsDB fetchExisting(long windowStartTime) throws Exception {
```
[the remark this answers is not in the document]
ricardolstephen: It would not. This method makes use of the MetricsDB ctor, which throws an Exception if there is some issue creating a connection to that file (like if the file is corrupt or something).

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -235,10 +255,39 @@ public void putMetric(Metric<Double> metric, Dimensions dimensions, long windowS
     return create.select(allFields).from(finalTable).groupBy(groupByFields).fetch();
   }
 
+  /**
+   * Queries all the data associated with the given metric.
+   *
+   * @param metric the desired metric
+   * @return the result of the query
+   */
   public Result<Record> queryMetric(String metric) {
```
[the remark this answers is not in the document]
ricardolstephen: That method is not part of this PR. Just some additional documentation was added. Will open a new ticket about it: opendistro-for-elasticsearch/performance-analyzer-rca#412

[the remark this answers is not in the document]
ricardolstephen: That method is not part of this PR. Just some additional documentation was added. Will open a new ticket about it: opendistro-for-elasticsearch/performance-analyzer-rca#412

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -235,10 +255,39 @@ public void putMetric(Metric<Double> metric, Dimensions dimensions, long windowS
...
 
+  /**
+   * Queries all the data associated with a given metric.
+   *
+   * @param metric the desired metric
+   * @param dimensions the dimensions we want to return for the given metric
+   * @param limit the maximum number of records to return
+   * @return the result of the query
+   */
+  public Result<Record> queryMetric(String metric, Collection<String> dimensions, int limit) {
```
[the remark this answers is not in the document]
ricardolstephen: Agreed.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -259,6 +308,23 @@ public void deleteOnDiskFile() {
...
   }
 
+  /**
+   * Deletes the metricsdb file associated with the given timestamp if it exists.
+   *
+   * @param windowStartTime the timestamp associated with an existing metricsdb file
+   */
+  public static void deleteOnDiskFile(long windowStartTime) {
+    String dbFilePath = getDBFilePath(windowStartTime);
+    File dbFile = new File(dbFilePath);
```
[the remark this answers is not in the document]
ricardolstephen: Such a check for the existence of the file is unnecessary -- perhaps after the check the file may not be there. That is why the delete operation itself takes care of all of that.
We should keep it's name as "deleteOnDiskFile" to stay consistent with current "deleteOnDiskFile". Both files to do the same thing, it's just this new version is static and accepts a windowStartTime argument.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -71,6 +82,11 @@
   private final AppContext appContext;
   private final ConfigOverridesApplier configOverridesApplier;
 
+  public static final String BATCH_METRICS_ENABLED_CONF_FILE = "batch_metrics_enabled.conf";
+  private boolean batchMetricsEnabled;
+  private boolean defaultBatchMetricsEnabled = false;
```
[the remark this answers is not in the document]
ricardolstephen: We should keep it as-is to remain consistent with rcaEnabledDefaultValue. Additinally, this value might change, so we shouldn't make it final.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -754,6 +828,28 @@ public void emitNodeMetrics(long currWindowStartTime, MetricsDB metricsDB) throw
     }
   }
 
+  @VisibleForTesting
+  public void readBatchMetricsEnabledFromConf() {
```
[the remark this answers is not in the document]
ricardolstephen: VisibleForTesting "Annotates a program element that exists, or is more widely visible than otherwise necessary, only for use in test code". This method would be private if we only needed it in trimOldSnapshots. However, we need it for testing purposes, so we make it VisibleForTesting.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -754,6 +828,28 @@ public void emitNodeMetrics(long currWindowStartTime, MetricsDB metricsDB) throw
...
+                String nextLine = sc.nextLine();
+                boolean oldValue = batchMetricsEnabled;
+                boolean newValue = Boolean.parseBoolean(nextLine);
+                if (oldValue != newValue) {
+                  batchMetricsEnabled = newValue;
+                  LOG.info("Batch metrics enabled changed from {} to {}", oldValue, newValue);
+                }
+              } catch (IOException e) {
+                LOG.error("Error reading file '{}': {}", filePath.toString(), e);
+                e.printStackTrace();
```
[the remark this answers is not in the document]
ricardolstephen: We should keep this as-is to stay consistent with readRcaEnabledFromConf.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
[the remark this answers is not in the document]
ricardolstephen: We should keep trimDatabases for now. It's been pretty well tested and it is public. The codebase itself can use some cleanup, so this can be taken care of with that. However, we should retain the code for now in case we want to make use of it down the road.

See https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/pull/315#discussion_r475793690

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: This is not what we do. We add just the timestamp to a set. Then, the query handler uses the timestamps from this set to call fetchExisting. fetchExisting creates an additional connection to an already existing database (we do not clone the snapshot), and we access the data via that connection.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: This call does belong here (see the comment above it). The method name does not need to be renamed -- the same abstraction is achieved at the end. Additionally, this is complexity that cannot be avoided. The comments are here to help explain the complexity to the reader.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: Sure

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
yojs: In no case we should be manually constructing json. Please use the Gson library. It will make it easier and less error prone. Maybe you can look at HotNodeSummary::toJson()

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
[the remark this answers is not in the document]
yojs: That might be misleading. That method should not be called from anywhere in the code other than tests. For your usecase, you can make that protected and then have a wrapper Test class that exposes it over a public interface for tests only ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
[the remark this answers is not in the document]
yojs: Right, so you don't need to throw a InvalidParameterException, you can simple return ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
```diff
@@ -271,4 +295,21 @@ private void loadMetricsDBFilesCleanupEnabled() {
...
+      }
+      batchMetricsRetentionPeriodMinutes = parsedRetentionPeriod;
+    } catch (NumberFormatException | InvalidParameterException e) {
+      LOG.error("Invalid batch-metrics-retention-period-minutes. Using default value {}.",
+              batchMetricsRetentionPeriodMinutes, e);
```
yojs, line 313: Do we need the stack trace here ? 
And as this is logged for both `NumberFormatException` and `InvalidParameterException`, it might make sense to print `(settings.getProperty(BATCH_METRICS_RETENTION_PERIOD_MINUTES)` to understand which of the two it was by just looking at the logs ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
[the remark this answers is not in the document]
yojs: In that case use Files.delete() and catch the exception if one is thrown  for understanding why it could not be deleted. The error message here is not helpful

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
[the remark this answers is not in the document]
yojs: Can you explain how can a default change ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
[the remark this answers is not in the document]
yojs: If we are removing the usage in the code, we should remove the method. Git takes care of preserving history. So, if it is required in future, we can resurrect it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
yojs: Sorry, for the confusion, by not belong here, I meant in terms of the name of the method. So please change the name to reflect that ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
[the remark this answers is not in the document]
yojs: One of the ways to avoid it, if we at all want to test private methods, would be to encapsulate all such methods, in a wrapper class of your own. Then make the methods inside the wrapper as public and hence testable. The wrapper can be used in the main class as private or its getters/setters as visible for testing.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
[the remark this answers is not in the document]
yojs: We may not want to do everything for consistency as much as we want to do the right thing. :) This may be redundant as the stack will already be captured in the logs.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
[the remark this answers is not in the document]
ricardolstephen: I am throwing an InvalidParameterException so I can catch it (along with the NumberFormatException that may be thrown by Long.parseLong) and log the error.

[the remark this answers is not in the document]
ricardolstephen: But yeah, it seems like it might be more appropriate to just give more user-friendly information about the value being out of range, like loadMetricsDeletionIntervalFromConfig is.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
ricardolstephen: Sure will get rid of the exception log and doing something like this.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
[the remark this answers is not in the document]
ricardolstephen: Sure

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -201,8 +219,46 @@ public void trimOldSnapshots() throws Exception {
     trimMap(shardRqMetricsMap, RQ_SNAPSHOTS);
     trimMap(httpRqMetricsMap, HTTP_RQ_SNAPSHOTS);
     trimMap(masterEventMetricsMap, MASTER_EVENT_SNAPSHOTS);
-    trimDatabases(
-        metricsDBMap, MAX_DATABASES, PluginSettings.instance().shouldCleanupMetricsDBFiles());
+
+    boolean deleteDBFiles = PluginSettings.instance().shouldCleanupMetricsDBFiles();
+    // Cleanup all but the 2 most recent metricsDB files from metricsDBMap. The most recent metricsDB files needs to be
+    // retained for future metrics query handling, the second most recent metricsDB file needs to be retained in case
```
yojs, line 225: We have a few things going on here:
- cleaning up in-memory snapshots
- cleaning up on-disk files if batch metrics is disabled.
- re-reading config 
- And cleaning up metrics db files so that we always have a fixed number of them around.

Let's move the in-memory cleanup and on disk cleanups into separate methods of their own

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -201,8 +219,46 @@ public void trimOldSnapshots() throws Exception {
...
+      if (deleteDBFiles) {
+        for (Long timestamp : batchMetricsDBSet) {
+          if (!metricsDBMap.containsKey(timestamp)) {
+            MetricsDB.deleteOnDiskFile(timestamp);
+          }
+        }
+      }
+      batchMetricsDBSet.clear();
+    }
+    readBatchMetricsEnabledFromConf();
```
yojs, line 251: Although we are playing safe here by not reading the config before, but I don't think we are completely avoiding the race here. Say, if reading all the files in the retention period is taking too long, (multiple iterations of this call), then there is still a chance that the API read thread tries to read a file that has been deleted. So, we should definitely handle that exception, in any case.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -696,6 +755,24 @@ public void deleteDBs() throws Exception {
...
+   * This is called by operations outside of the ReaderMetricsProcessor.
+   *
+   * @return A list of the timestamps associated with MetricsDB files. The oldest of the MetricsDB files typically
+   *     have a lifetime of ~SAMPLING_INTERVAL seconds (no less than SAMPLING_INTERVAL/2 seconds). Null if batch
+   *     metrics is disabled.
+   */
+  public NavigableSet<Long> getBatchMetrics() {
+    if (batchMetricsEnabled) {
+      TreeSet<Long> batchMetricsDBSetCopy = new TreeSet<>(batchMetricsDBSet.clone());
+      while (batchMetricsDBSetCopy.size() > PluginSettings.instance().getBatchMetricsRetentionPeriodMinutes() * 12) {
```
yojs, line 768: `PluginSettings.instance().getBatchMetricsRetentionPeriodMinutes() * 12`

This is a call we are making on each iteration of the loop. Can we store it in a local variable as it does not change over the loop iterations ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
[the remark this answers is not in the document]
ricardolstephen: Sure, will just remove it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -71,6 +82,11 @@
   private final AppContext appContext;
   private final ConfigOverridesApplier configOverridesApplier;
 
+  public static final String BATCH_METRICS_ENABLED_CONF_FILE = "batch_metrics_enabled.conf";
+  private boolean batchMetricsEnabled;
+  private boolean defaultBatchMetricsEnabled = false;
+  private ConcurrentSkipListSet<Long> batchMetricsDBSet;
```
yojs, line 88: Let's add a comment here saying that this datastructure can be modified by the writer/cleanup thread while being read by the other thread trying to batch metrics. And therefore, it needs to be `ConcurrentSkipListSet`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
```diff
@@ -0,0 +1,309 @@
...
+
+  private static final Logger LOG = LogManager.getLogger(QueryBatchRequestHandler.class);
+
+  private static final int TIME_OUT_VALUE = 2;
+  private static final TimeUnit TIME_OUT_UNIT = TimeUnit.SECONDS;
+  private NetClient netClient;
+  MetricsRestUtil metricsRestUtil;
+
+  public static final int DEFAULT_MAX_DATAPOINTS = 100800;  // Must be non-negative
+  public static final long DEFAULT_SAMPLING_PERIOD_MILLIS = 5000;  // Must be a multiple of 5000
```
yojs, line 88: Let's also add the reason for the numbers here ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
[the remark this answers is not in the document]
ricardolstephen: Sure, will remove it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -71,6 +82,11 @@
...
+
```
yojs, line 89: `batchMetricsDBSet` holds the key to the castle and it is an in-memory structure. In case PA starts crashing say every 15 seconds, or so. So, we will not be cleaning up files that were generated in the last run as this structure is always initialized to an empty set !

yojs (changes requested):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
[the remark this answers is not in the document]
ricardolstephen: When I mentioned "this value might change", I meant that we may choose to make it true by default. But yeah, since this comment, my notion of constants has changed. Will change it to be private static final.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
ricardolstephen: Max datapoints was capped to prevent excessive memory consumption. Will note that in README and design docs. However, I don't see the need to note anything about DEFAULT_SAMPLING_PERIOD_MILLIS.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
```diff
@@ -0,0 +1,309 @@
...
+        samplingPeriod *= 1000;
+      }
+
+      if (startTime >= endTime) {
+        throw new InvalidParameterException("starttime must be less than the endtime");
+      }
+      startTime -= startTime % samplingPeriod;
+      endTime -= endTime % samplingPeriod;
+      if (startTime == endTime) {
+        throw new InvalidParameterException("starttime and endtime cannot be equal when rounded down to the nearest sampling period");
```
vigyasharma, line 196: nit: should we reword as something on the lines of - "Start time and end time should be at least sampling period apart".? It may be confusing in its current form.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
```diff
@@ -0,0 +1,309 @@
...
+      }
+      if (endTime > currentTime) {
+        throw new InvalidParameterException("endtime can be no greater than the system time at the node");
```
vigyasharma, line 199: Minor: System time at node should not matter, right? Since startTime and endTime are provided as utc epoch timestamp values. We can just mention something like expected value is utc seconds since epoch timestamp and provided value is in future.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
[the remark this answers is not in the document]
vigyasharma: What I meant was that `if (maxDatapoints == 0)` is a very exact check. We might want `<=` instead?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
vigyasharma: I agree with joydeep and sid. There are standard tested libraries that we should leverage. Can move to a separate PR as a follow up.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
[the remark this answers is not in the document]
vigyasharma: > it's just this new version is static and accepts a windowStartTime argument.

why do we need a separate static instance of the same method, instead of overloading `public void deleteOnDiskFile()` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: Sure

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: Sure

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
ricardolstephen: Agreed

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
ricardolstephen: I use the phrase "system time" because we determine current time using System.currentTimeMillis, which in turn uses methods from the operating system. And I believe machines can be setup to indicate whatever timestamp the owner wants -- like, I could setup a computer today with a current time of January 1, 1970. So, I expose this additional detail in the error message to help anyone who might be having problems, while keeping the readme and design doc at the abstraction of utc for the standard use case.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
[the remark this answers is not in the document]
ricardolstephen: Logically, maxDatapoints should never be less than zero. But, sure.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: This is true. Will file this as a bug.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rest/QueryBatchRequestHandler.java
ricardolstephen: Sure, will raise a new enhancement issue about it #416

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: Right, the metrics processor code gives the query handler's a ~5s grace period before deleting stale metricsdb files (see the javadoc on getBatchMetrics). For the typical case (when batch metrics is not about to be disabled), this will not be an issue since the batch metrics query handler reads data from the older files first. For the case when batch metrics is about to be disabled, even newer files are at risk of deletion. However, this disabling of batch metrics is not expected to be a use case, there is a 100,800 datapoints limit on the query handler, there's not much data to say that this grace period will be an issue given the datapoint limit, and the query handler properly handles missing files.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: Sure

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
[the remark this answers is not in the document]
ricardolstephen: Will address this by making the target methods private and exposing them via VisibleForTesting shims.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: So, if the reader keeps crashing every 15 seconds, there will not be much impact since the rmp waits for a I believe ~30s or longer before actually producing metricsDB files. However, I believe the issue you mention actually raises the bigger concern that metricsdb files are never cleaned up -- whether they are collected for metrics or batch metrics. As a result, if the reader crashes with the same periodicity as the metricsdb retention period (10s with just metrics processing and no batch metrics, or a peak of 60min with batch metrics) then the cluster will just accumulate a large number of metricsdb files that aren't cleaned up.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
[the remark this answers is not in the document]
ricardolstephen: Will address this by making the target methods private and exposing them via VisibleForTesting shims.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: Sure, will separate out into a trimOldSnapshots method and a trimOldMetricsDBFiles method.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: For this initial pr, we can just cleanup any files that were found. However, we will only close ticket #415 once we've implemented initializing the batch metrics set with the available files found on startup.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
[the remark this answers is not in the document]
ricardolstephen: There's two separate abstractions we want. 1. Given a MetricsDB instance, we want to be able to delete it's on-disk file (the non-static version). 2. We want to be able to able to delete arbitrary metrcsdb files without creating an instance (the static version). Creating an instance would have the cost of setting up a sql connection to it... which itself could actually fail (that's why the ctor throws an Exception).

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java
```diff
@@ -327,6 +331,36 @@ public static void deleteOnDiskFile(long windowStartTime) {
...
+   *
+   * @return the timestamps associated with on-disk files
+   */
+  public static Set<Long> listOnDiskFiles() {
+    String prefix = PluginSettings.instance().getSettingValue(DB_FILE_PREFIX_PATH_CONF_NAME, DB_FILE_PREFIX_PATH_DEFAULT);
+    Path prefixPath = Paths.get(prefix);
+    Path parentPath = prefixPath.getParent();
+    Set<Long> found = new HashSet<Long>();
+    try (Stream<Path> paths = Files.list(parentPath)) {
+      PathMatcher matcher = FileSystems.getDefault().getPathMatcher("regex:" + prefix + "\\d+");
```
yojs, line 345: We might skip the overhead of regex compiling by stripping off the `parentPath` from the `prefix` -  the remainder is what we are looking for in the directory.

yojs (dismissed):

vigyasharma (approved):

sidheart (approved):

yojs (approved):

PR #323, opened 2020-07-30
Persistance concurrency bug

yojs:
*Issue #:* #324 

*Description of changes:*
This change makes us print less stacktrace on errors that are expected. Also resolves a concurrency bug while threads doing writes and while the file is being rotated.

*Tests:*
Added a test


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
```diff
@@ -98,37 +106,40 @@ synchronized void createTable(String tableName, List<Field<?>> columns, String r
...
       jooqTableColumns.put(tableName, columns);
-    } catch (Exception e) {
-      LOG.error("Failed to create table {}", tableName);
-      throw new SQLException();
+    } catch (DataAccessException e) {
+      String msg = "table " + tableName + " already exists";
+      if (!e.getMessage().contains(msg)) {
+        LOG.error(e);
+        throw new SQLException(e);
+      }
```
ktkrg, line 125: Since we are consuming the exception, can we add a log statement for the else case that we were not expecting the table to have been created already but we're going to use the one already created?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/FileRotate.java
```diff
@@ -115,7 +116,14 @@ protected Path rotate(long currentMillis) throws IOException {
       Files.move(FILE_TO_ROTATE, targetFilePath);
       lastRotatedMillis = System.currentTimeMillis();
     } catch (FileAlreadyExistsException fae) {
-      LOG.error(fae);
+      if (!Files.deleteIfExists(targetFilePath)) {
+        LOG.error("Could not delete file: " + targetFilePath);
+      }
+      try {
+        Files.move(FILE_TO_ROTATE, targetFilePath, StandardCopyOption.REPLACE_EXISTING);
```
ktkrg, line 123: We should add StandardCopyOption.ATOMIC_MOVE to the copy options as well.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
yojs: good point !

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/FileRotate.java
yojs: 👍

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
```diff
@@ -98,37 +111,54 @@ synchronized void createTable(String tableName, List<Field<?>> columns, String r
...
-        .column(DSL.field(getPrimaryKeyColumnName(tableName) + PRIMARY_KEY_AUTOINCREMENT_POSTFIX))
-        .columns(columns)
-        .constraints(DSL.constraint(foreignKeyField.getName() + "_FK").foreignKey(foreignKeyField)
-            .references(referenceTable, DSL.field(referenceTablePrimaryKeyFieldName)));
 
-    LOG.debug("table with fk created: {}", constraintStep.toString());
     try {
+      LOG.debug("Trying to create a summary table: {} that references {}", tableName, referenceTableName);
+      Table referenceTable = DSL.table(referenceTableName);
+      CreateTableConstraintStep constraintStep = create.createTable(tableName)
```
rguo-aws, line 118: I assume creating the SQL statement itself does not throw any exception. Can we leave this part out of the try block ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
```diff
@@ -98,37 +111,54 @@ synchronized void createTable(String tableName, List<Field<?>> columns, String r
...
+    Objects.requireNonNull(create, "DSLContext cannot be null");
+    Table<Record> table = DSL.table(tableName);
+    List<Field<?>> columnsForTable = jooqTableColumns.get(tableName);
+    if (columnsForTable == null) {
+      LOG.error("NO columns found for table: {}. Tables: {}, columns: {}", tableName, tableNames, jooqTableColumns);
+      throw new SQLException("No columns exist for table.");
+    }
+
     try {
+      InsertValuesStepN insertValuesStepN = create
```
rguo-aws, line 152: same as above

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
yojs: The DSLContext can be uninitialized if we are not using synchronization between the threads. I saw a case where the table was added to hastable but the columns for the table were missing. We gather some date to create the createStatement and it the gathering part we can run into exceptions.
These cases are fixed now but it would be good to catch and log exception. But you are right, the scope of exceptions is larger and so I have added code to catch exception and not just DataAcessException.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
yojs: This create statement is less involved and therefore changes of exceptions are less. Will move it out.

PR #325, opened 2020-07-31
Reader changes for dynamic enable/disable of RCA graph components

### LINKED ISSUE #327: Support dynamic enable/disabling of RCA, decider nodes and actions.
We need a mechanism to dynamically enable and disable various RCAs, deciders and the recommended actions.

ktkrg:
*Issue #:*
Fixes #327 

*Description of changes:*
This change adds functionality to read the overrides written by the NodeDetailsCollector and update the rca conf file accordingly. To that effect this change adds two new sections to the `rca[_idle][_master].conf`:
`mutedActions: []`
`mutedDeciders: []`

Updating RCA conf and the muted nodes work for RCAs and deciders. For actions, the list of muted actions are looked at by the collator and prunes the ones that are muted.

*Tests:*
Ran tests with different overrides and made sure that the rca conf was updated correctly and muted correctly. Verified same things for the actions as well.


vigyasharma:
General comment on the description message:
> For actions, the list of muted actions are looked at by the collator and prunes the ones that are muted.

Should this be done at collator level or decider level? I was thinking that if an action is muted, it returns `isActionable() == false`, and decider picks the next best possible action. If we mute it at collator level, then decider will keep picking that same muted action.

ktkrg:
> General comment on the description message:
> 
> > For actions, the list of muted actions are looked at by the collator and prunes the ones that are muted.
> 
> Should this be done at collator level or decider level? I was thinking that if an action is muted, it returns `isActionable() == false`, and decider picks the next best possible action. If we mute it at collator level, then decider will keep picking that same muted action.

Aah yes, I started this before I started on collator, so I missed the fact that deciders can suggest multiple actions based.

Will change this.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
```diff
@@ -0,0 +1,117 @@
...
+            currentMutedRcaSet.removeAll(overrides.getEnable().getRcas());
+          }
+          if (overrides.getEnable().getDeciders() != null) {
+            currentMutedDeciderSet.removeAll(overrides.getEnable().getDeciders());
+          }
+          if (overrides.getEnable().getActions() != null) {
+            currentMutedActionSet.removeAll(overrides.getEnable().getActions());
+          }
+        }
+
```
khushbr, line 79: Do we need DEBUG level log statements here for the components removed and similarly for below, added to muted components?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -319,26 +319,66 @@ private void readRcaEnabledFromConf() {
...
+          LOG.error("Incorrect RCA(s): {}, cannot be muted. Valid RCAs: {}, Muted RCAs: {}",
+              rcaConf.getMutedRcaList(), ConnectedComponent.getNodeNames(),
+              Stats.getInstance().getMutedGraphNodes());
+          return false;
+        }
+      }
+
+      LOG.info("Updating the muted graph nodes to : {}", graphNodesForMute);
+      Stats.getInstance().updateMutedGraphNodes(graphNodesForMute);
+      Stats.getInstance().updateMutedActions(actionsForMute);
```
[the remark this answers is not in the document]
ktkrg: No, we don't have a registry of available actions to check the input against. As such, it'll just stay in the conf if an invalid string is provided.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/Stats.java
```diff
@@ -29,6 +30,7 @@
   private Map<Integer, ConnectedComponent> graphs;
 
   private Set<String> mutedGraphNodes;
+  private Set<String> mutedActions;
```
yojs, line 33: It would be nice if we don't add code to this class anymore and in future, we just deprecate it. Adding anything to this class restricts us to have a uniform view of things across all virtualInstances during RCA-IT and we won't be able to test things like nodes running different versions of graphs and different muted rca as such.

we have this new kid in the block to pass around things - `appContext`. Any chance we can use it ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
```diff
@@ -75,15 +87,18 @@ public void processEvent(Event event) {
 
     // An example node_metrics data is something like this for a two node cluster:
     // {"current_time":1566414001749}
+    // 1566414001749
     // {"overrides": {"enabled": {}, "disabled": {}}
```
yojs, line 91: This is not valid json: we are missing a `}` in the end

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
```diff
@@ -0,0 +1,117 @@
...
+  public void applyOverride(final String overridesJson, final String lastUpdatedTimestampString) {
+    if (!valid(overridesJson, lastUpdatedTimestampString)) {
+      LOG.warn("Received invalid overrides or timestamp. Overrides json: {}, last updated "
+          + "timestamp: {}", overridesJson, lastAppliedTimestamp);
+      return;
+    }
+
+    try {
+      long lastUpdatedTimestamp = Long.parseLong(lastUpdatedTimestampString);
```
yojs, line 45: Do we want to handle parse exception ?

src/test/resources/rca/rca_elected_master.conf
```diff
@@ -73,6 +73,8 @@
   },
 
   "muted-rcas": "",
```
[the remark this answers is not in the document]
ktkrg: Done.

src/test/resources/rca/rca.conf
```diff
@@ -63,6 +63,8 @@
   },
 
   "muted-rcas": "",
```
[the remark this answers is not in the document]
ktkrg: Done.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCTest.java
```diff
@@ -17,6 +17,7 @@
 import org.apache.logging.log4j.Logger;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
```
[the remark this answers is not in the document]
ktkrg: removed.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesWrapper.java
```diff
@@ -15,13 +15,29 @@
...
+
+    public ConfigOverridesWrapper() {
+        this(new ObjectMapper());
+    }
+
+    /**
+     * Ctor used only for unit test purposes.
+     * @param mapper The object mapper instance.
+     */
+    public ConfigOverridesWrapper(final ObjectMapper mapper) {
```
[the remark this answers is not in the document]
ktkrg: Done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -56,7 +56,7 @@ public String name() {
   }
 
   @Override
-  public boolean isActionable() {
+  public boolean isValid() {
```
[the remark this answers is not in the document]
ktkrg: I liked the previous one better too, the problem was isActionable() makes sense for an action and in this case I wanted to enforce the muted-ness check in the isActionable() so that isActionable() can return false if the action is muted. The only way to do that is to override the method in an abstract class and make the method final and provide a hook for the action implementations to tell if they can perform some updates to the system(what they were doing in the original isActionable() method).

Changed it to canUpdate().

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -319,26 +319,66 @@ private void readRcaEnabledFromConf() {
         });
   }
 
-  private void readAndUpdateMutesRcasDuringStart() {
+  private void readAndUpdateMutedRcasDuringStart() {
```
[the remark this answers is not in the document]
ktkrg: Done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -319,26 +319,66 @@ private void readRcaEnabledFromConf() {
...
+        }
+      }
+
+      LOG.info("Updating the muted graph nodes to : {}", graphNodesForMute);
+      Stats.getInstance().updateMutedGraphNodes(graphNodesForMute);
+      Stats.getInstance().updateMutedActions(actionsForMute);
     } catch (Exception e) {
-      LOG.error("Couldn't read/update the muted RCAs during start()", e);
+      LOG.error("Couldn't read/update the muted RCAs", e);
       StatsCollector.instance().logMetric(RCA_MUTE_ERROR_METRIC);
```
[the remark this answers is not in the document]
ktkrg: Updated to MUTE_ERROR_METRIC.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/ConfJsonWrapper.java
```diff
@@ -44,6 +45,10 @@
   private final List<String> mutedRcaList;
```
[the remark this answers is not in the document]
ktkrg: Done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
ktkrg: Added.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    if (overridesJson.isEmpty() || timestamp.isEmpty()) {
+      return false;
+    }
+
+    return NumberUtils.isCreatable(timestamp);
+  }
+
+  public long getLastAppliedTimestamp() {
```
yojs, line 114: The callers of this are the test methods. Let's add a `VisibleForTesting` annotation.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
```diff
@@ -0,0 +1,117 @@
...
+      }
+    } catch (IOException ioe) {
+      LOG.error("Unable to deserialize overrides JSON:" + overridesJson, ioe);
+    }
+  }
+
+  private void apply(final ConfigOverrides overrides) {
+    if (PerformanceAnalyzerApp.getRcaController() != null && PerformanceAnalyzerApp.getRcaController().isRcaEnabled()) {
+      LOG.info("Applying overrides: {}", overrides.getEnable().getRcas());
+      RcaConf rcaConf = PerformanceAnalyzerApp.getRcaController().getRcaConf();
```
yojs, line 62: Should we get the list of rca.conf from the RcaControllerHelper and apply to all three of them ? If the node role changes, then the muting will be invalidated.

Any changes that you make to the static set of objects in PerformanceAnalyzerApp may not be tested using RCA-IT.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/RcaConf.java
```diff
@@ -155,21 +187,56 @@ public CacheDeciderConfig getCacheDeciderConfig() {
...
     return setting;
   }
 
+  public void updateRcaConf(final Set<String> mutedRcas, final Set<String> mutedDeciders,
+      final Set<String> mutedActions) {
+    String updatedPath = this.configFileLoc + ".updated";
+    try (final FileInputStream originalFileInputStream = new FileInputStream(this.configFileLoc);
+        final Scanner scanner = new Scanner(originalFileInputStream, StandardCharsets.UTF_8.name());
+        final FileOutputStream updatedFileOutputStream = new FileOutputStream(updatedPath)
+    ) {
```
yojs, line 210: can we move this to the line above ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
```diff
@@ -41,7 +42,15 @@
...
   private volatile ImmutableList<NodeDetails> nodesDetails = null;
 
-  public ClusterDetailsEventProcessor() {}
+  private final ConfigOverridesApplier overridesApplier;
+
+  public ClusterDetailsEventProcessor() {
+    this(new ConfigOverridesApplier());
+  }
+
+  public ClusterDetailsEventProcessor(final ConfigOverridesApplier overridesApplier) {
```
yojs, line 51: Maybe we don't need to pass `ConfigOverridesApplier` as a constructor. No state is kept in it. It reads the config and applies it to the `rca.conf`. Maybe we can just use it as a helper static class ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.logging.log4j.Logger;
+
+/**
+ * Class the manages applying the various overrides for RCAs, deciders, and action nodes as
+ * written by the NodeDetails collector and read by the {@link ClusterDetailsEventProcessor}
+ */
+public class ConfigOverridesApplier {
+
+  private static final Logger LOG = LogManager.getLogger(ConfigOverridesApplier.class);
+  private volatile long lastAppliedTimestamp;
```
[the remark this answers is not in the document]
ktkrg: removed.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
```diff
@@ -75,15 +87,18 @@ public void processEvent(Event event) {
 
     // An example node_metrics data is something like this for a two node cluster:
     // {"current_time":1566414001749}
+    // 1566414001749
```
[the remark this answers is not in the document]
ktkrg: This needs a change in the plugin as well. Will follow up with a separate PR for this.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/Stats.java
ktkrg: yup, using it now.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
ktkrg: Added.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
ktkrg: Handling now.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
ktkrg: done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
ktkrg: Good catch! updating all conf files now.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/RcaConf.java
ktkrg: done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
ktkrg: It does keep the `lastAppliedTimestamp` as an instance variable and uses that to check if the config overrides string received from the `ClusterDetailsEventProcessor`. Making it a static helper will make all the consumers see the same value of last applied time(in an IT for example).

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ClusterDetailsEventProcessor.java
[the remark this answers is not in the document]
yojs: cool. But let's change it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
```diff
@@ -0,0 +1,127 @@
...
+            currentMutedDeciderSet.addAll(overrides.getDisable().getDeciders());
+          }
+          if (overrides.getDisable().getActions() != null) {
+            currentMutedActionSet.addAll(overrides.getDisable().getActions());
+          }
+        }
+
+        LOG.debug("New set of muted rcas: {}", currentMutedRcaSet);
+        LOG.debug("New set of muted deciders: {}", currentMutedDeciderSet);
+        LOG.debug("New set of muted actions: {}", currentMutedActionSet);
```
yojs, line 99: We should be applying change when someone calls the API or at startup, which shouldn't be that many. Should we make it info logs ? Or add a metric.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
```diff
@@ -0,0 +1,127 @@
...
+        rcaConf.updateAllRcaConfFiles(currentMutedRcaSet, currentMutedDeciderSet, currentMutedActionSet);
```
yojs, line 100: Now we are updating three `rca.conf` files. `updateAllRcaConfFiles` has a void return type. What happens if it succeeded on a subset of files but not all? We will still update `lastAppliedTimestamp` and never look back until someone makes a change again. So, should we make `lastAppliedTimestamp` return true when writes on all three files succeed and and then only update the `lastAppliedTimestamp`. If not, then we will retry again on the next run and eventually all three files will be updated or we likely have a disk corruption or disk full issues. thoughts ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/AppContext.java
```diff
@@ -39,10 +40,12 @@
   // initiate a node config cache within each AppContext space
   // to store node config settings from ES
   private final NodeConfigCache nodeConfigCache;
+  private Set<String> mutedActions;
```
yojs, line 43: The RCAController thread updates this but it is read by the SchedulerTask threads. We should make this `volatile`

ktkrg: Done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/overrides/ConfigOverridesApplier.java
ktkrg: That's a good point! I'm checking all three updates before updating the lastAppliedTimestamp.

I'm also emitting a metric if we fail to update an rca conf file. If we end up in a state where we can't do these file system changes, I think we should be failing in more places than just this, so yea, it's probably ok to do the timestamp update only if all three succeed.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/RcaConf.java
```diff
@@ -202,17 +203,24 @@ public CacheDeciderConfig getCacheDeciderConfig() {
...
     // update all rca.conf files
     List<String> rcaConfFiles = RcaControllerHelper.getAllConfFilePaths();
-
     for (String confFilePath : rcaConfFiles) {
-      updateRcaConf(confFilePath, mutedRcas, mutedDeciders, mutedActions);
+      updateStatus = updateRcaConf(confFilePath, mutedRcas, mutedDeciders,
+        mutedActions);
+      if (!updateStatus) {
+        LOG.error("Failed to update the conf file at path: {}", confFilePath);
+        StatsCollector.instance().logMetric(RcaConsts.WRITE_UPDATED_RCA_CONF_ERROR);
```
yojs, line 216: You can use the new metric framework. There your metric can also tell you which file erred out ?

PR #326, opened 2020-07-31
JVM decider

rguo-aws:
*Issue #:*

*Description of changes:*
Add JVM decider to RCA framework

*Tests:*
WIP


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
```diff
@@ -0,0 +1,217 @@
...
+
+  // allocate value to its bucket
+  private int bucketization(int lowerBound, int upperBound, int val, int bucketSize) {
+    double step = (double) (upperBound - lowerBound) / (double) bucketSize;
+    return (int) ((double) val / step);
+  }
+
+  // downsize queue based on priority and current queue size
+  private void actionPriorityForQueue() {
+    Integer writeQueueEWMASize = NodeConfigCacheUtil.readQueueSize(esNode, nodeConfigCache, ResourceEnum.WRITE_THREADPOOL);
```
yojs, line 136: What is `EWMA` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
```diff
@@ -0,0 +1,217 @@
...
+    Integer searchQueueEWMASize = NodeConfigCacheUtil.readQueueSize(esNode, nodeConfigCache, ResourceEnum.SEARCH_THREADPOOL);
+    if (writeQueueEWMASize == null || searchQueueEWMASize == null) {
```
yojs, line 138: Would we not want to take some actions if only one of the queues is non-null ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelOneActionBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+    actionFilter.put(ResourceEnum.FIELD_DATA_CACHE, true);
+    actionFilter.put(ResourceEnum.SHARD_REQUEST_CACHE, true);
+  }
+
+  /**
+   * build actions for level 1 bucket.
+   * @return List of actions
+   */
+  public List<Action> buildActions() {
+    addFieldDataCacheAction();
```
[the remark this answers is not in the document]
rguo-aws: done. moved to constructor

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
```diff
@@ -0,0 +1,217 @@
...
+    caches.add(ResourceEnum.FIELD_DATA_CACHE);
+    caches.add(ResourceEnum.SHARD_REQUEST_CACHE);
+    List<ResourceEnum> queues = new ArrayList<>();
+    queues.add(ResourceEnum.WRITE_THREADPOOL);
+    queues.add(ResourceEnum.SEARCH_THREADPOOL);
+    targetCaches = Collections.unmodifiableList(caches);
+    targetQueues = Collections.unmodifiableList(queues);
+  }
+
+  public LevelTwoActionBuilder(final NodeKey esNode, final NodeConfigCache nodeConfigCache) {
```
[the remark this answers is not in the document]
rguo-aws: done. created a base class to maintain common code

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/OldGenDecisionPolicy.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public List<Action> actions(final NodeKey esNode, double oldGenUsage) {
+    if (oldGenUsage >= OLD_GEN_THRESHOLD_LEVEL_THREE) {
+      return LevelThreeActionBuilder.newBuilder(esNode, nodeConfigCache).build();
+    }
+    else if (oldGenUsage >= OLD_GEN_THRESHOLD_LEVEL_TWO) {
+      return LevelTwoActionBuilder.newBuilder(esNode, nodeConfigCache).build();
+    }
+    else if (oldGenUsage >= OLD_GEN_THRESHOLD_LEVEL_ONE) {
+      return LevelOneActionBuilder.newBuilder(esNode, nodeConfigCache).build();
```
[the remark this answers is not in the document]
sidheart: After rereading this my thoughts are the same. I don't understand what problem we were trying to solve by using the builder pattern here. We don't actually use the builders as legitimate builders, they're just wrappers around specific logic for generating a list of actions given a known bucket level. Why can't the static variables be held in this class with the List<Action> generated in actions() itself? Who would reuse these classes down the line?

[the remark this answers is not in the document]
sidheart: The only moderately complex logic seems to live entirely in LevelTwoActionBuilder#actionPriorityForQueue(), but might not this just live in a function in this class?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/BaseActionBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+  protected abstract void actionPriorityFilter();
+
+  /**
+   * build actions.
+   * @return List of actions
+   */
+  public List<Action> build() {
+    List<Action> actions = new ArrayList<>();
+    cacheActionMap.forEach((cache, action) -> {
+      if (actionFilter.containsKey(cache) && actionFilter.get(cache)) {
```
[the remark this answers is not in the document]
rguo-aws: thanks, that is a good point.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/BaseActionBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+        actions.add(action);
+      }
+    });
+    queueActionMap.forEach((queue, action) -> {
+      if (actionFilter.containsKey(queue) && actionFilter.get(queue)) {
```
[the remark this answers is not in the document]
rguo-aws: done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelOneActionBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+    Long capacity = NodeConfigCacheReaderUtil
+        .readCacheMaxSizeInBytes(nodeConfigCache, esNode, ResourceEnum.FIELD_DATA_CACHE);
+    if (capacity == null) {
+      return;
+    }
+    ModifyCacheMaxSizeAction action = new ModifyCacheMaxSizeAction(esNode, ResourceEnum.FIELD_DATA_CACHE, nodeConfigCache,
+        capacity, false, LEVEL_ONE_CONST.CACHE_ACTION_STEP_COUNT);
+    if (action.isActionable()
+        && action.getDesiredCapacityInPercent() >= LEVEL_ONE_CONST.FIELD_DATA_CACHE_LOWER_BOUND) {
+      cacheActionMap.put(ResourceEnum.FIELD_DATA_CACHE, action);
```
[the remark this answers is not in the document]
rguo-aws: JVM decider doesn't really care about lower bound because the end goal is to decrease cache/queue capacity and here we want to make sure the new capacity proposed by this action is above the lower bound set in this level

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelThreeActionBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * build actions if old gen falls into level three bucket
+ *
+ * <p>if old gen usage(after full gc) falls into this bucket 90% - 100%, JVM heap is extremely
+ * contended and can run into OOM at any moment. So action builder will build a group of actions
+ * to scale down caches to their lower bound in one shot. And for queues we will downsize all
+ * queues simultaneously with even higher steps
+ */
+public class LevelThreeActionBuilder extends BaseActionBuilder {
```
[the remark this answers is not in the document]
rguo-aws: I am also trying to figure out if there is any better name here. But the Level* here does not correspond to the usage of JVM memory.  That is because the threshold of JVM healthiness is 60%(used to be 65%) and in JVM decider we further divide the range of unhealthiness (60%-100%) into three subranges

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/OldGenDecisionPolicy.java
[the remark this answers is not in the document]
rguo-aws: You are right, the complex logic right now only sit in Level two but as I mention above, we need to collect more data from experiment and then make changes based on that. So now level 1 & 3 are more like placeholder with very simple logics in it and each bucket in JVM decider can be evolved independently in the future.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelOneActionBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+  /**
+   * generate final action list based on action priority.
+   * The default priority in this level is to downsize both caches simultaneously
+   * unless explicitly overridden by customer yml.
+   * @return final action list based on action priority
+   */
+  // TODO : read priority from yml if customer wants to override default ordering
+  @Override
+  protected void actionPriorityFilter() {
+    actionFilter.put(ResourceEnum.FIELD_DATA_CACHE, true);
```
vigyasharma, line 90: How does this ensure a priority order? Keys in the `HashMap` are not ordered, right? I also didn't understand the significance of `boolean` value field

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelThreeActionBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.NodeKey;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.util.NodeConfigCacheReaderUtil;
+
+/**
+ * build actions if old gen falls into level three bucket
+ *
+ * <p>if old gen usage(after full gc) falls into this bucket 90% - 100%, JVM heap is extremely
+ * contended and can run into OOM at any moment. So action builder will build a group of actions
+ * to scale down caches to their lower bound in one shot. And for queues we will downsize all
+ * queues simultaneously with even higher steps
```
vigyasharma, line 31: Do you also want to add a `cacheClearAction` for level 3, that clears all ES caches, but runs with a high cool off time to avoid repeated invocations?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelOneActionBuilder.java
vigyasharma: IIUC, you are using the `actionFilter` to check which actions to pick based on different levels? e.g. only the actions allowed at that level will be marked `true`. Is this understanding correct? 
If so, then why don't we just create only the relevant actions in each level and do away with this action filter.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelThreeActionBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  //downsize field data cache to its lower bound in one shot
+  public void addFieldDataCacheAction() {
+    Long capacity = NodeConfigCacheReaderUtil
+        .readCacheMaxSizeInBytes(nodeConfigCache, esNode, ResourceEnum.FIELD_DATA_CACHE);
+    if (capacity == null) {
+      return;
+    }
+    ModifyCacheMaxSizeAction action = ModifyCacheMaxSizeAction.newMinimalCapacityAction(esNode,
```
vigyasharma, line 50: Minor: This can be done via a flag in the builder now.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+        .readQueueEWMASize(nodeConfigCache, esNode, ResourceEnum.WRITE_THREADPOOL);
+    Integer searchQueueEWMASize = NodeConfigCacheReaderUtil
+        .readQueueEWMASize(nodeConfigCache, esNode, ResourceEnum.SEARCH_THREADPOOL);
+    if (writeQueueEWMASize == null || searchQueueEWMASize == null) {
+      return;
+    }
+    ModifyQueueCapacityAction writeQueueAction = queueActionMap.get(ResourceEnum.WRITE_THREADPOOL);
+    ModifyQueueCapacityAction searchQueueAction = queueActionMap.get(ResourceEnum.SEARCH_THREADPOOL);
+    if (writeQueueAction != null && searchQueueAction != null) {
+      int writeQueueSizeBucket = bucketization(writeQueueAction.getLowerBound(),
```
vigyasharma, line 125: I see what you are doing with `bucketization` here, this is good. We should add a doc string to explain to future readers. 

The important point being that we cannot just look at `{w|s}QueueEWMASize/{w|s}QueueCapacity` because if `QueueCapacity` is already very low, say 200 for write queue, and all of it is constantly full (likely with small queues), we will get %age used as ~100% and keep reducing the same small queue. 

Hence we need to look at the full range of sizes available for each queue (lower bound to upper bound), to get a sense of which queue is using more mem. We also don't simply compare absolute values of `{w|s}QueueEWMASize`s because they inherently have different overheads. Comparing them according to their full range ensures we maintain the same proportion while comparing them.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+
+  private void actionPriorityForCache() {
+    actionFilter.put(ResourceEnum.FIELD_DATA_CACHE, true);
+    actionFilter.put(ResourceEnum.SHARD_REQUEST_CACHE, true);
+  }
+
+  // allocate value to its bucket
+  private int bucketization(int lowerBound, int upperBound, int val, int bucketSize) {
```
vigyasharma, line 108: Add a doc string to explain this formulae, and what `step` means here

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+    actionPriorityForQueue();
+  }
+
+  //TODO : read consts from rca.conf
+  private static class LEVEL_TWO_CONST {
+    public static final double FIELD_DATA_CACHE_LOWER_BOUND = 0.02;
+    public static final double SHARD_REQUEST_CACHE_LOWER_BOUND = 0.01;
+    public static final int CACHE_ACTION_STEP_COUNT = 2;
+    public static final int QUEUE_ACTION_STEP_COUNT = 1;
+    public static final int QUEUE_SIZE_BUCKET_SIZE = 10;
```
vigyasharma, line 181: Should this be a separate constant? These are just bands that you are using to divide the range of queue capacity, can this just be local to the `bucketization` function?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/OldGenDecisionPolicy.java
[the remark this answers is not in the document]
vigyasharma: I agree with Sid, I don't really see a need for the common base class here. We have different strategies for each level that we want to be able to evolve independently. A simple interface might be enough.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/OldGenDecisionPolicy.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  public List<Action> actions(final NodeKey esNode, double oldGenUsage) {
+    if (oldGenUsage >= OLD_GEN_THRESHOLD_LEVEL_THREE) {
+      return LevelThreeActionBuilder.newBuilder(esNode, nodeConfigCache).build();
+    }
+    else if (oldGenUsage >= OLD_GEN_THRESHOLD_LEVEL_TWO) {
+      return LevelTwoActionBuilder.newBuilder(esNode, nodeConfigCache).build();
+    }
+    else if (oldGenUsage >= OLD_GEN_THRESHOLD_LEVEL_ONE) {
```
vigyasharma, line 48: Do you want to give more *observation* time for level 2 and level 3? Level 2 & 3 are post GC IHOP - should we factor in some time for gc trying and failing to clear things up?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelOneActionBuilder.java
rguo-aws: The intention of creating an action filter is to de-couple the action generation part away from the priority ordering in conf file. By doing this we can avoid a very large function which adds actions and re-order actions based on rca.conf all together in one place.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/OldGenDecisionPolicy.java
rguo-aws: Even the oldGenUsage for level 1 is post GC usage I think we should factor in more time for this entire oldGenDecisonPolicy. Do you think that should be done via cool off period or should be done at RCA scheduler level ? 
One approach is we keeps the decider running but set the cool off period of all queue/cache actions to a very large number (more than 10mins)
The other is we change the decisionFrequency in heap decider to run old gen decision logic at a much lower freq. 

What do you think ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
rguo-aws: done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/BaseActionBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+    this.cacheActionMap = new HashMap<>();
+    this.queueActionMap = new HashMap<>();
+    this.actionFilter = new HashMap<>();
+    registerActions();
+    actionPriorityFilter();
+  }
+
+  protected abstract void registerActions();
+
+  protected abstract void actionPriorityFilter();
```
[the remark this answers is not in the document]
rguo-aws: removed base class

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
rguo-aws: done. Created a ThreadPoolConfig object to store default config settings read from rca.conf. Use the lower - upper bounds defined in rca.conf (or the default values ) to calculate bucket size.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -40,14 +41,23 @@
   private Map<ResourceEnum, Integer> lowerBound = new HashMap<>();
   private Map<ResourceEnum, Integer> upperBound = new HashMap<>();
 
+  public ModifyQueueCapacityAction(NodeKey esNode, ResourceEnum threadPool, int currentCapacity, boolean increase, int step) {
+    this(esNode, threadPool, currentCapacity);
+    int desiredCapacity = increase ? currentCapacity + step * STEP_SIZE : currentCapacity - step * STEP_SIZE;
```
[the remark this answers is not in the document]
rguo-aws: done. we can now pass full step size as parameter in action builder

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -144,25 +147,24 @@ public ResourceEnum getCacheType() {
...
     private boolean canUpdate;
     private long coolOffPeriodInMillis;
 
     private Long currentCacheMaxSizeInBytes;
-    private Long desiredCacheMaxSizeInBytes;
+    private Double desiredCacheMaxSizeInPercent;
     private Long heapMaxSizeInBytes;
 
-    private Builder(
+    public Builder(
```
vigyasharma, line 161: Please review and rebase changes from #402 once merged

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
```diff
@@ -87,6 +91,11 @@ public Decision operate() {
     return decision;
   }
 
+  @Override
+  public void readRcaConf(RcaConf conf) {
+    deciderConfig = conf.getDeciderConfig();
+  }
+
```
vigyasharma, line 98: This should already be present in the `Decider` base class

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelOneActionBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+  DeciderConfig deciderConfig) {
+    return new LevelOneActionBuilder(esNode, appContext, deciderConfig);
+  }
+
+  private void addFieldDataCacheAction() {
+    ModifyCacheMaxSizeAction action = ModifyCacheMaxSizeAction
+        .newBuilder(esNode, ResourceEnum.FIELD_DATA_CACHE, appContext)
+        .increase(false)
+        .lowerBoundThreshold(actionBuilderConfig.fieldDataCacheLowerBound())
+        .upperBoundThreshold(cacheBoundConfig.fieldDataCacheUpperBound())
```
vigyasharma, line 74: Update with changes in #402

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelOneActionBuilder.java
vigyasharma: I see. But from current changes, I don't see any action being added to `actionFilter` as `false`, so it seems like we just add actions to the `actionFilter` map as true, and pick all of them in `build` for each `level*ActionBuilder`. What am i missing?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/decider/CacheBoundConfig.java
```diff
@@ -0,0 +1,73 @@
...
+ * express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.configs.decider;
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.configs.ConfigUtils;
+import java.util.Map;
+
+public class CacheBoundConfig {
```
vigyasharma, line 21: Let's use changes in #402 ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+
+  private void actionPriorityForCache() {
+    actionFilter.put(ResourceEnum.FIELD_DATA_CACHE, true);
+    actionFilter.put(ResourceEnum.SHARD_REQUEST_CACHE, true);
+  }
+
+  /**
+   * This function divide the range {lower bound - upper bound } of search/wrire queue into
```
vigyasharma, line 136: typo: write

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/decider/jvm/LevelOneActionBuilderConfig.java
```diff
@@ -0,0 +1,32 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.configs.decider.jvm;
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.configs.ConfigUtils;
+import java.util.Map;
+
+public class LevelOneActionBuilderConfig {
+  public static final double DEFAULT_FIELD_DATA_CACHE_LOWER_BOUND = 0.1;
+  public static final double DEFAULT_SHARD_REQUEST_CACHE_LOWER_BOUND = 0.02;
+  private static String FIELD_DATA_CACHE_LOWER_BOUND_CONFIG_NAME = "fielddata-cache-lower-bound";
+  private static String SHARD_REQUEST_CACHE_LOWER_BOUND_CONFIG_NAME = "shard-request-cache-lower-bound";
```
vigyasharma, line 10: Let's move these to `CacheActionConfig`? we can call them `soft-upper/lower-bound`.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/decider/jvm/LevelTwoActionBuilderConfig.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Map;
+
+public class LevelTwoActionBuilderConfig {
+  public static final double DEFAULT_FIELD_DATA_CACHE_LOWER_BOUND = 0.02;
+  public static final double DEFAULT_SHARD_REQUEST_CACHE_LOWER_BOUND = 0.01;
+  public static final int DEFAULT_CACHE_ACTION_STEP_COUNT = 2;
+  public static final int DEFAULT_QUEUE_ACTION_STEP_COUNT = 1;
+  public static final int DEFAULT_QUEUE_BUCKET_SIZE = 10;
+  private static String FIELD_DATA_CACHE_LOWER_BOUND_CONFIG_NAME = "fielddata-cache-lower-bound";
+  private static String SHARD_REQUEST_CACHE_LOWER_BOUND_CONFIG_NAME = "shard-request-cache-lower-bound";
```
vigyasharma, line 13: Pls see if you really need a **lower-bound, level-1-bound, level-2-bound, upper-bound**? Can we do away with level-2 bound?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/decider/jvm/LevelTwoActionBuilderConfig.java
```diff
@@ -0,0 +1,48 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.configs.decider.jvm;
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.configs.ConfigUtils;
+import java.util.Map;
+
+public class LevelTwoActionBuilderConfig {
```
vigyasharma, line 6: Can these be merged with queue/cache action configs?

rguo-aws: this config object currently defines the step size of each action that decider is going to issue in each level. and we might add more knobs for each level in the future. so i think it would be better to keep this config separately

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/configs/CacheActionConfig.java
```diff
@@ -64,9 +68,24 @@ public CacheActionConfig(RcaConf conf) {
...
+  }
+
+  /**
+   * this function calculate the size of a single step given the range {lower bound - upper bound}
+   * and number of steps
+   */
+  public double getStepSize(ResourceEnum cacheType) {
+    ThresholdConfig<Double> threshold = getThresholdConfig(cacheType);
+    return (threshold.upperBound() - threshold.lowerBound()) / (double) getTotalStepCount();
+  }
```
vigyasharma, line 87: Nice!

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
vigyasharma: @rguo-aws pls fix before merging

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
```diff
@@ -0,0 +1,247 @@
...
+      if (writeQueueSizeBucket > searchQueueSizeBucket) {
+        actionFilter.put(ResourceEnum.WRITE_THREADPOOL, true);
+      }
+      else if (writeQueueSizeBucket < searchQueueSizeBucket) {
+        actionFilter.put(ResourceEnum.SEARCH_THREADPOOL, true);
+      }
+      // tie breaker
+      else {
+        if (workLoadTypeConfig.preferIngestOverSearch()) {
+          actionFilter.put(ResourceEnum.WRITE_THREADPOOL, true);
```
vigyasharma, line 190: if ingest is preferable, we should downsize search queue, no?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/decider/CachePriorityOrderConfig.java
```diff
@@ -0,0 +1,45 @@
...
+  private static final String PRIORITY_ORDER_CONFIG_NAME = "priority-order";
+  private static String FIELDDATA_CACHE = "fielddata-cache";
+  private static String SHARD_REQUEST_CACHE = "shard-request-cache";
+  public static final List<String> DEFAULT_PRIORITY_ORDER = Collections.unmodifiableList(
+      Arrays.asList(FIELDDATA_CACHE, SHARD_REQUEST_CACHE));
+  private Config<List<String>> priorityOrder;
+
+  public CachePriorityOrderConfig(NestedConfig configs) {
+    priorityOrder = new Config(PRIORITY_ORDER_CONFIG_NAME, configs.getValue(),
+        DEFAULT_PRIORITY_ORDER, List.class);
```
vigyasharma, line 39: Does this work? I was having trouble casting with generics

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/decider/WorkLoadTypeConfig.java
```diff
@@ -0,0 +1,66 @@
...
+  private static final String PRIORITY_ORDER_CONFIG_NAME = "priority-order";
+  private static String INGEST = "ingest";
+  private static String SEARCH = "search";
+  public static final List<String> DEFAULT_PRIORITY_ORDER = Collections.unmodifiableList(
+      Arrays.asList(INGEST, SEARCH));
+  private Config<List<String>> priorityOrder;
+  private Predicate<List<String>> listValidator;
+
+  public WorkLoadTypeConfig(NestedConfig configs) {
+    listValidator = (list) -> {
```
vigyasharma, line 41: Since there is only search and ingest (2 values), we can make this simpler by renaming the config. Instead of taking a list, we can do things like:
```
{ "decider-config-settings": {
    "prefer-ingest-over-search": true
    }
}
```

vigyasharma (commented):
Looks good. Will you add tests in a separate PR? If so, can you create an issue to track it?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/decider/WorkLoadTypeConfig.java
rguo-aws: done. changed the config to 
"workload-type": {
      "prefer-ingest": true,
      "prefer-search": false
    },

This will allow user to not assign workload preference to any queue type

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/configs/CacheActionConfig.java
sidheart: +1

sidheart (approved):
lgtm thanks for cleaning this up

vigyasharma (approved):

PR #328, opened 2020-07-31
Rca bucketization

sidheart:
*Issue #:*

*Description of changes:*

*Tests:*


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/HotResourceSummary.java
```diff
@@ -60,6 +60,16 @@
   private String metaData;
   private List<TopConsumerSummary> topConsumerSummaryList;
 
+  private UsageBucket usageBucket;
```
rguo-aws, line 63: HotResourceSummary is shared among various RCAs and is closed coupled with rpc / SQL persistor so we don't want to add extra variable unless it is necessary. Since we are emitting bucket information from cluster level RCA, can we define a new type of flowunit and add bucket object into the flowunit itself ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/JvmUsageBucketThresholds.java
```diff
@@ -0,0 +1,26 @@
...
+  private StaticBucketThresholds oldGenHeapUsageThresholds;
+
+  public JvmUsageBucketThresholds(StaticBucketThresholds youngGenPromotionRateThresholds,
+      StaticBucketThresholds oldGenHeapUsageThresholds) {
+    this.youngGenPromotionRateThresholds = youngGenPromotionRateThresholds;
+    this.oldGenHeapUsageThresholds = oldGenHeapUsageThresholds;
+  }
+
+  @Override
+  public UsageBucket computeBucket(HotResourceSummary resourceSummary) {
```
rguo-aws, line 16: same as above. I think this needs to be moved into flowunit

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/HighHeapUsageClusterRca.java
```diff
@@ -48,133 +42,91 @@
...
-      if (unhealthyNodeList.size() > 0) {
-        context = new ResourceContext(Resources.State.UNHEALTHY);
-        summary = new HotClusterSummary(
-            getAllClusterInstances().size(),
-            unhealthyNodeList.size());
-        for (HotNodeSummary unhealthyNodeSummary : unhealthyNodeList) {
-          summary.appendNestedSummary(unhealthyNodeSummary);
+      // youngGenSummaries can have multiple elements but we will only consider it as unhealthy if
+      // three consecutive summaries are all unhealthy and we will then pick the first element as the summary for output.
+      if (youngGenSummaries.size() >= UNHEALTHY_FLOWUNIT_THRESHOLD
```
rguo-aws, line 127: This actually comes from legacy code and I think it might be unnecessary to keep track of most recent three flowunits anymore. Let's keep this JVM cluster RCA simple by only keeping track of single flowunit form each node

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/HotResourceSummary.java
sidheart: Moved into ClusterFlowUnit and the associated classes

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/JvmUsageBucketThresholds.java
sidheart: Moved into the ClusterFlowUnit class

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/HighHeapUsageClusterRca.java
sidheart: Fixed

pa_config/rca.conf
```diff
@@ -97,5 +97,11 @@
       "field-data-cache-upper-bound" : 0.4,
       "shard-request-cache-upper-bound" : 0.05
     }
+  },
+
+  "usage-bucket-thresholds": {
```
vigyasharma, line 102: Some doc string on how to interpret these config values will be useful

pa_config/rca.conf
```diff
@@ -97,5 +97,11 @@
...
+    "young-gen-heap-promotion-rate": [100.0, 250.0, 400.0],
+    "old-gen-heap-usage": [20.0, 40.0, 75.0],
```
vigyasharma, line 104: Are these %ages. For other such values (like upper/lower bound), these have been expressed as decimals e.g. `0.2 for 20%`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/flow_units/ClusterResourceFlowUnit.java
```diff
@@ -0,0 +1,99 @@
...
+   * <p>You probably don't need to override this. Just implement initBucketCalculator()
+   */
+  protected void computeUsageBuckets() {
+    if (getSummary() == null) {
+      return;
+    }
+    for (HotNodeSummary hotNodeSummary : getSummary().getHotNodeSummaryList()) {
+      for (HotResourceSummary hotResourceSummary : hotNodeSummary.getHotResourceSummaryList()) {
+        NodeKey nodeKey = new NodeKey(hotNodeSummary.getNodeID(), hotNodeSummary.getHostAddress());
+        if (bucketCalculator != null) {
```
vigyasharma, line 75: If we dont do anything when `bucketCalculator == null` then we can move this check to the beginning of the fn.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/flow_units/ClusterResourceFlowUnit.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Base class for cluster level {@link ResourceFlowUnit}s which want to bucketize resources into
+ * various {@link UsageBucket}s
+ *
+ * <p>If you'd like to subclass this class, simply implement initBucketCalculator() to return an
+ * appropriate {@link BucketCalculator} to use for bucketization. {@link CpuClusterResourceFlowUnit}
+ * is a good example of this.
+ */
+public class ClusterResourceFlowUnit extends ResourceFlowUnit<HotClusterSummary> {
```
vigyasharma, line 36: There are too many levels of inheritance happening here - `GenericFlowUnit -> ResourceFlowUnit -> ClusterResourceFlowUnit -> {Cpu/Jvm}ClusterResourceFlowUnit`. Can some of it be moved to composition? Like an interface for `BucketizedFlowUnit` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/flow_units/ClusterResourceFlowUnit.java
```diff
@@ -0,0 +1,99 @@
...
+   * Returns the {@link UsageBucket} that a resource was placed into for a given node or
+   * {@link UsageBucket#UNKNOWN} if this value was never computed
+   *
+   * @param nodeKey The node whose resource we're interested in
+   * @param resource The actual resource we're interested in
+   * @return the {@link UsageBucket} that a resource was placed into for a given node or
+   *         {@link UsageBucket#UNKNOWN} if this value was never computed
+   */
+  public UsageBucket getUsageBucket(NodeKey nodeKey, ResourceEnum resource) {
+    if (bucketCalculator == null) {
```
vigyasharma, line 94: Let's log an error message here?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/flow_units/CpuClusterResourceFlowUnit.java
```diff
@@ -0,0 +1,47 @@
...
+      RcaConf rcaConf,
+      boolean persistSummary) {
+    super(timeStamp, context, summary, rcaConf, persistSummary);
+  }
+
+  @Override
+  public BucketCalculator initBucketCalculator() {
+    if (rcaConf == null) {
+      return null;
+    }
```
vigyasharma, line 42: Defaults should be present in the class itself, and `rca.conf` should only contain overrides if any.

pa_config/rca.conf
```diff
@@ -97,5 +97,11 @@
       "field-data-cache-upper-bound" : 0.4,
       "shard-request-cache-upper-bound" : 0.05
     }
+  },
+
+  "usage-bucket-thresholds": {
+    "young-gen-heap-promotion-rate": [100.0, 250.0, 400.0],
```
vigyasharma, line 103: Thinking it would be more user friendly to have a key-value pair for describing these thresholds, rather than positional variables. This current design requires both users and developers to remember what each position stands for. And it is not future proof to if you add more buckets.
Something like:
```ruby
"young-gen-heap-promotion-rate": {
    "under-utilized": 20.0,
    "healthy-with-buffer": 100.0, 
    "healthy": 250.0, 
    "unhealthy": 400.0
}
```

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/flow_units/JvmClusterResourceFlowUnit.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @Override
+  public BucketCalculator initBucketCalculator() {
+    if (rcaConf == null) {
+      return null;
+    }
+    List<Double> youngGenHeapPromotionRateThresholds = rcaConf.getUsageBucketThresholds(BucketCalculator.YOUNG_GEN_PROMOTION_RATE);
+    List<Double> oldGenHeapUsageThresholds = rcaConf.getUsageBucketThresholds(BucketCalculator.OLD_GEN_HEAP_USAGE);
+    BucketCalculator youngGenCalculator = new StaticBucketCalculator(
```
vigyasharma, line 47: As mentioned above, let's decouple from positional significance of these values

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/AnyClusterSummary.java
```diff
@@ -0,0 +1,36 @@
...
+ *  permissions and limitations under the License.
+ */
+
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries;
+
+/**
+ * AnyNodeSummary is a HotClusterSummary which may contain a summary of healthy values
+ */
+public class AnyClusterSummary extends HotClusterSummary {
+  private boolean isHot;
```
vigyasharma, line 22: Did you mean `AnyNode` or `AnyCluster` summary? why do we need a new class, can we not rename the existing classes?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/HotClusterSummary.java
```diff
@@ -59,6 +59,10 @@ public HotClusterSummary(int numOfNodes, int numOfUnhealthyNodes) {
     this.hotNodeSummaryList = new ArrayList<>();
   }
 
+  public void setNumOfUnhealthyNodes(int numOfUnhealthyNodes) {
```
vigyasharma, line 62: nit: rename to `totalUnhealthyNodes` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/bucket/BucketCalculator.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * A BucketCalculator identifies which {@link UsageBucket} a {@link Resource} should be placed
+ * into given that resource's value. It does this on a per-node basis.
+ *
+ * <p>E.g. a BucketCalculator can compute a value of HEALTHY for CPU on Node "A" and a value of
+ * HEALTHY_WITH_BUFFER for CPU on Node "B". A consumer of this information can then read out these
+ * bucket values by calling something like getUsageBucket(NodeA, CPU).
+ */
+public abstract class BucketCalculator {
```
vigyasharma, line 34: Can this be an interface instead?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/bucket/BucketCalculator.java
```diff
@@ -0,0 +1,86 @@
...
+  private final Logger LOG = LogManager.getLogger(getClass());
+  // Strings which map to threshold definitions in rca.conf
+  public static final String YOUNG_GEN_PROMOTION_RATE = "young-gen-heap-promotion-rate";
+  public static final String OLD_GEN_HEAP_USAGE = "old-gen-heap-usage";
+  public static final String CPU_USAGE = "cpu-usage";
+
+  // For each Node, store a map of resource name to the UsageBucket for that resource
+  protected Map<NodeKey, Map<ResourceEnum, UsageBucket>> resourceBucketMap = new HashMap<>();
+
+  public void computeBucket(NodeKey nodeKey, ResourceEnum resource, double value) {
```
vigyasharma, line 44: Does `BucketCalculator` need to store what was the computed bucket for each resource on each node? How about we just let ClusterLevelRCAs store this info, and only use BucketCalculators to compute what the appropriate bucket is?
Right now the computed state for an RCA is spread across too many classes, which in themselves have multi-level abstract and concrete classes. It's getting hard to track.
I've not gotten to the end of the PR, doing it this way may be the best approach. But adding a note for you to reconsider if any changes can make things easier to follow.

sidheart:
Breaking this up into smaller PRs

PR #330, opened 2020-08-01
Adding Cache RCAs to Analysis Graph, using cache max size from Node Config Cache and minor fixes

khushbr:
*Description of changes:*
1. Adding the Cache Node and Cluster RCAs to Analysis graph
2. Enabling query on Cache Cluster RCAs via RCA API
3. Minor Re-factoring in CacheUtil
4. Updating to use Cache Max size from Node Config Cache
```

[ec2-user@ip-10-212-22-125 logs]$ curl --url "localhost:9650/_opendistro/_performanceanalyzer/rca?name=ShardRequestCacheClusterRca" -XGET | python -m json.tool
{
    "ShardRequestCacheClusterRca": [
        {
            "HotClusterSummary": [
                {
                    "HotNodeSummary": [
                        {
                            "HotResourceSummary": [
                                {
                                    "avg": NaN,
                                    "max": NaN,
                                    "meta_data": "",
                                    "min": NaN,
                                    "resource_metric": "cache eviction(cache eviction count)",
                                    "resource_type": "shard request cache",
                                    "threshold": 300.0,
                                    "time_period_seconds": 0,
                                    "value": 455.0
                                }
                            ],
                            "host_address": "10.212.22.125",
                            "node_id": "RtStKpfBSkqN1U8nkvDxSw"
                        }
                    ],
                    "number_of_nodes": 2,
                    "number_of_unhealthy_nodes": 1
                }
            ],
            "rca_name": "ShardRequestCacheClusterRca",
            "state": "unhealthy",
            "timestamp": 1596247532956
        }
    ]
}
```

*Tests:* Unit Tests and Dev Stack Testing


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/CacheUtil.java
```diff
@@ -15,44 +15,64 @@
...
             if (flowUnit.getData().size() > 0) {
-                double size = flowUnit.getData().get(0).getValue(MetricsDB.SUM, Double.class);
-                if (Double.isNaN(size)) {
-                    LOG.error("Failed to parse metric in FlowUnit from {}", cacheSizeGroupByOperation.getClass().getName());
-                } else {
-                    sizeTotalInKB += size / 1024.0;
-                }
+                Result<Record> records = flowUnit.getData();
+                double size = records.stream().mapToDouble(
+                        record -> record.getValue(MetricsDB.SUM, Double.class)).sum();
```
vigyasharma, line 46: Can any of these Record values be NaN?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/CacheUtil.java
```diff
@@ -15,44 +15,64 @@
...
             }
         }
-        return sizeTotalInKB;
+        return totalSizeInKB;
+    }
+
+    public static Double getSizeInKB(double sizeinBytes) {
+        double sizeInKB = 0;
+        if (Double.isNaN(sizeinBytes)) {
+            LOG.error("getSizeInKB called with NaN value");
```
vigyasharma, line 56: Can we throw an exception here instead? I try to avoid overloading return values with error codes in java. Off the top of my head, if cache size ever is less than 1KB, we cannot distinguish between that and error values.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/CacheUtil.java
```diff
@@ -15,44 +15,64 @@
...
+        } else {
+            sizeInKB = sizeinBytes / 1024.0;
+        }
+        return sizeInKB;
+    }
+
+    public static double getCacheMaxSize(AppContext appContext, NodeKey esNode, Resource cacheResource) {
+        try {
+            return appContext.getNodeConfigCache().get(esNode, cacheResource);
```
vigyasharma, line 65: Again, why don't we just throw the exception and let callers handle it their way. If we want to prevent it from breaking code, we should throw a `checked` exception here.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElasticSearchAnalysisGraph.java
```diff
@@ -203,6 +212,58 @@ public void construct() {
...
+            MetricsDB.MAX, ShardStatsDerivedDimension.INDEX_NAME.toString());
+    shardRequestCacheSizeGroupByOperation.addTag(TAG_LOCUS, LOCUS_DATA_MASTER_NODE);
+    addLeaf(shardRequestCacheSizeGroupByOperation);
+
+    ShardRequestCacheRca shardRequestCacheNodeRca = new ShardRequestCacheRca(RCA_PERIOD,
+            shardRequestCacheEvictions,
+            shardRequestHits,
+            shardRequestCacheSizeGroupByOperation);
+    shardRequestCacheNodeRca.addTag(TAG_LOCUS, LOCUS_DATA_MASTER_NODE);
+    shardRequestCacheNodeRca.addAllUpstreams(Collections.singletonList(shardRequestHits));
```
sruti1312, line 258: shardRequestCacheNodeRca needs to be connected to shardRequestCacheEvictions leaf node as well right?
Arrays.asList(All the metrics this node is consuming)?

khushbr: Thank You for catching this issue, we need to add remainder metrics as well.  Adding them.

PR #332, opened 2020-08-03
Polarize actions based on impact vectors

### LINKED ISSUE #331: Collator should polarize actions based on impact vectors
The current implementation of the collator is a pass-through implementation.

The collator should be able to take the actions suggested by various deciders and group them by node. Then it should also ensure that the impact vectors for each node is aligned in one direction(i.e. it should be polarized), either in the direction of increasing pressure on the node or decreasing pressure on the node.

In the case of a conflict where there are multiple actions suggested by different deciders that will both increase the pressure and decrease the pressure, the collator should prune the pressure increasing action as it should favor stability(decreasing pressure) over performance(increasing pressure).

ktkrg:
*Issue #:*
Fixes #331 

*Description of changes:*
This change adds the ability for the collator to group decider actions by nodes and polarize the actions based on their impact vectors.

The collator looks at various actions suggested by the deciders and first groups them by the node those actions impact.

For each node, if there is an action suggested that would go in the direction of decreasing pressure, then only choose(a la polarize) pressure decreasing actions. Only if there is no pressure decreasing action suggested for a node, then choose actions that increase pressure.

*Tests:*
Unit tests and tested on docker.


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * <p>Decisions can increase or decrease pressure on different key resources on an Elasticearch
+ * node. This is encapsulated in each Action via the {@link ImpactVector}. Since each decider
+ * independently evaluates its decision, it is possible to have conflicting ImpactVectors from
+ * actions across deciders.
+ *
+ * <p>The collator prunes them to ensure we only take actions that either increase, or decrease
+ * pressure on a particular node. To resolve conflicts, we prefer stability over performance.
+ */
+public class Collator extends Decider {
```
yojs, line 42: There can only be one collator node in a graph right ? Can we add that to the comments ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
```diff
@@ -0,0 +1,125 @@
...
+        List<Decision> decisions = decider.getFlowUnits();
+        if (decisions != null) {
+          decisions.forEach(decision -> proposedActions.addAll(decision.getActions()));
+        }
+      }
+    }
+    return proposedActions;
+  }
+
+  private List<Action> polarize(final NodeKey nodeKey, List<Action> actions) {
```
yojs, line 96: maybe we can make actions final as well ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
```diff
@@ -0,0 +1,125 @@
...
+
+    for (final Action action : actions) {
+      ImpactVector impactVector = action.impact().getOrDefault(nodeKey, new ImpactVector());
+
+      // Classify the action as pressure increasing action if the impact for any dimension is
+      // increasing pressure.
+      if (impactVector.getImpact()
+                      .values()
+                      .stream()
+                      .anyMatch(impact -> impact == Impact.INCREASES_PRESSURE)) {
```
yojs, line 108: Functional programming !! yay !

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/SingleNodeImpactActionGrouper.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Override
+  @NonNull public Map<NodeKey, List<Action>> groupByNodeId(@NonNull List<Action> actions) {
+    final Map<NodeKey, List<Action>> actionsByNodeId = new HashMap<>();
+    actions.stream()
+           .filter(action -> action.impactedNodes().size() == 1)
+           .forEach(action -> actionsByNodeId.computeIfAbsent(action.impactedNodes()
+                                                                    .get(0), k -> new ArrayList<>())
+                                             .add(action));
+    return actionsByNodeId;
```
yojs, line 36: Do we need to turn on some checkstyle rukes, this should have caused an error ?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/SingleNodeImpactActionGrouperTest.java
```diff
@@ -0,0 +1,73 @@
...
+        testActionGrouper
+            .groupByNodeId(ImmutableList.of(singleNodeImpactAction1, multiNodeImpactAction1));
+
+    assertEquals(1, groupedActions.keySet().size());
+    assertTrue(groupedActions.containsKey(nodeKey1));
+    assertFalse(groupedActions.containsKey(nodeKey2));
+    assertEquals(1, groupedActions.get(nodeKey1).size());
+    assertEquals(singleNodeImpactAction1, groupedActions.get(nodeKey1).get(0));
+  }
+}
```
yojs, line 73: A new line in the end ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
```diff
@@ -0,0 +1,125 @@
...
+        pressureIncreasingActions.add(action);
+      } else {
+        pressureNonIncreasingActions.add(action);
```
khushbr, line 111: given "If there are any actions that decrease pressure for a node, prefer that over list of actions that increase pressure.", should we break here when we find a `pressureNonIncreasingActions` ?

yojs:
Can you add a label for the PR ?

yojs:
In the PR description 
>first groups them by the node

Unfortunately, `node` is an overloaded term which can mean a graph node or a cluster node. Can we clarify ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/SingleNodeImpactActionGrouper.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Map;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+public class SingleNodeImpactActionGrouper implements ActionGrouper {
+
+  @Override
+  @NonNull public Map<NodeKey, List<Action>> groupByNodeId(@NonNull List<Action> actions) {
+    final Map<NodeKey, List<Action>> actionsByNodeId = new HashMap<>();
+    actions.stream()
+           .filter(action -> action.impactedNodes().size() == 1)
```
khushbr, line 32: so, each action can be associated with 1 or 0 impacted Nodes ? Just wanted to clarify this.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/ActionGrouper.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public interface ActionGrouper {
+
+  /**
+   * Groups the given list of actions by the nodes they impact..
+   *
+   * @param actions The list of actions that need to be grouped.
+   * @return A map of actions grouped by nodes they impact.
+   */
+  @NonNull Map<NodeKey, List<Action>> groupByNodeId(@NonNull final List<Action> actions);
```
yojs, line 36: Can we rename the method name to include `InstanceName` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
```diff
@@ -0,0 +1,125 @@
...
+   * unnecessary delays. As soon as a decision is taken, it should be evaluated and published downstream.
+   */
+  private static final int collatorFrequency = 1; // Measured in terms of number of evaluationIntervalPeriods
+
+  private final List<Decider> deciders;
+
+  private final ActionGrouper actionGrouper;
+
+  public Collator(long evalIntervalSeconds, Decider... deciders) {
+    this(evalIntervalSeconds, new SingleNodeImpactActionGrouper(), deciders);
```
yojs, line 57: We plan to remove the `evalIntervalSeconds` as we wanted the scheduler to be simple - it would call each graphNode at every tick and then the node should decide how to handle if this is before the periodicity. In that vein, can we add a constant in `Node` class and use that and not even take `evalIntervalSeconds` as a constructor parameter ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
```diff
@@ -0,0 +1,125 @@
...
+    this.actionGrouper = actionGrouper;
+  }
+
+  @Override
+  public String name() {
+    return NAME;
+  }
+
+  @Override
+  public Decision operate() {
```
yojs, line 72: I think it might be a good idea to add a javadoc comment for the `operate()` as this is the center piece of the RCAGraph node ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/SingleNodeImpactActionGrouper.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+public class SingleNodeImpactActionGrouper implements ActionGrouper {
+
+  @Override
+  @NonNull public Map<NodeKey, List<Action>> groupByNodeId(@NonNull List<Action> actions) {
```
yojs, line 29: So, we want the collator to come up with one action per node or one action per resource per node ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/SingleNodeImpactActionGrouper.java
yojs, line 32: Can we add a comment why are we filtering out the actions that ask for the same action to be performed on multiple nodes ? Is that rationale that we do it only one node per iteration ?

ktkrg: Interesting point, don't we create multiple actions in such cases?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/SingleNodeImpactActionGrouper.java
ktkrg: Collator does not filter beyond polarizing the actions. It will return as many as possible after making sure that they align in terms of the impact on the instance so that the publisher and other components can filter based on other parameters.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
ktkrg: Added.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/ActionGrouper.java
ktkrg: yup, renamed it to `groupByInstanceId`.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/SingleNodeImpactActionGrouper.java
ktkrg: Yeah, is that not the understanding today? This class only considers those actions that impact exactly one node. The multi node impact action grouping will be done in the next PR.

@yojs brought up a similar point, do we create a single action with multiple nodes in the impacted nodes when we want to say decrease cache size for node1, node2, node3 or do we create multiple actions for each node we want to decrease the cache size for?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/SingleNodeImpactActionGrouper.java
ktkrg: the .get(0) is aligned with .impactedNodes() because they're chained calls, and similarly, .add(action) is aligned with .computeIfAbsent() because they're chained calls.

This is what the formatter did and I thought it was helpful 😅 I'm using the GoogleStyle scheme. Let me know if I should use another formatter.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/SingleNodeImpactActionGrouperTest.java
ktkrg: Added.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/ActionGrouper.java
```diff
@@ -0,0 +1,37 @@
...
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * A grouping interface to provide different ways to group actions as needed by the {@link
+ * Collator}
+ */
+public interface ActionGrouper {
+
+  /**
+   * Groups the given list of actions by the nodes they impact..
```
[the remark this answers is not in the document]
ktkrg: Done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
```diff
@@ -0,0 +1,125 @@
...
+    return finalDecision;
+  }
+
+  @NonNull private List<Action> getAllProposedActions() {
+    final List<Action> proposedActions = new ArrayList<>();
+    if (deciders != null) {
+      for (final Decider decider : deciders) {
+        List<Decision> decisions = decider.getFlowUnits();
+        if (decisions != null) {
+          decisions.forEach(decision -> proposedActions.addAll(decision.getActions()));
```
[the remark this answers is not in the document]
ktkrg: not sure if any actions return `null` values, but added a null check any way.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
```diff
@@ -0,0 +1,125 @@
...
+  public String name() {
+    return NAME;
+  }
+
+  @Override
+  public Decision operate() {
+    List<Action> proposedActions = getAllProposedActions();
+    Map<NodeKey, List<Action>> actionsByNode = actionGrouper.groupByNodeId(proposedActions);
+    List<Action> prunedActions = new ArrayList<>();
+    actionsByNode.forEach((k, v) -> prunedActions.addAll(polarize(k, v)));
```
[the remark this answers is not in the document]
ktkrg: Done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
ktkrg: there might be other pressure decreasing actions beyond the one that increases it, we need to add that as well, so we can't break early.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
ktkrg: Removed `evalIntervalSeconds` ctor parameter in Collator.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
ktkrg: Done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
ktkrg: done. Took care of the other parameters as well.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/SingleNodeImpactActionGrouper.java
khushbr: I clarified this with others and we will have action which impact a single node and then those which are across multiple nodes.

Taking the above example, cache increase action will be per node but for cache decrease, we will have 1 action with multiple nodes in the `impactedNodes`.

ktkrg: Looks like we are creating one action per node even if its the same action for multiple nodes, added a comment.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
```diff
@@ -0,0 +1,148 @@
...
+  private List<Action> polarize(final NodeKey nodeKey, final List<Action> actions) {
+    final List<Action> pressureIncreasingActions = new ArrayList<>();
+    final List<Action> pressureNonIncreasingActions = new ArrayList<>();
+
+    for (final Action action : actions) {
+      ImpactVector impactVector = action.impact().getOrDefault(nodeKey, new ImpactVector());
+
+      // Classify the action as pressure increasing action if the impact for any dimension is
+      // increasing pressure.
+      if (impactVector.getImpact()
```
vigyasharma, line 128: This classifies an increase across any dimension as increasing pressure. Shouldn't we check on a per dimension basis? Otherwise there is no value in having different dimensions in the impact vector.

Suppose you have an action to offload some data from heap to a file on disk. This would reduce heap but increase disk pressure. If there is no other action trying to decrease disk pressure, then the overall disk on the node is not under contention and we should let this action go through.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
```diff
@@ -0,0 +1,148 @@
...
+              proposedActions.addAll(decision.getActions());
+            }
+          });
+        }
+      }
+    }
+    return proposedActions;
+  }
+
+  private List<Action> polarize(final NodeKey nodeKey, final List<Action> actions) {
```
vigyasharma, line 119: It seems to me that this will need non-trivial changes when we have actions impacting multiple nodes. If an action is purged because of pressure conflict on one of the nodes, it must be removed from the list of all of the nodes. 

Should `polarize()` then move to `ActionGrouper`? Is the overall above structure right for future use cases?

---

My ideas for solving this were more around compiling all impact vectors into a matrix and doing a second pass through it to accept or prune actions. Something on these lines:

#### Suppose the collator received 2 actions, moveShard from node A to node B, and moveShard from node B to node C
```ruby
actions = [MoveShard: A -> B, MoveShard: B -> C]
Impact vectors:
	actions[0] :: MoveShard: A -> B
	impactedNodes = [A, B]
	A: [CPU:dec, Heap: dec, RAM: dec, NW: dec, Disk: dec]
	B: [CPU:inc, Heap: inc, RAM: inc, NW: inc, Disk: inc]

	actions[1] :: MoveShard: B -> C
	impactedNodes = [B, C]
	B: [CPU:dec, Heap: dec, RAM: dec, NW: dec, Disk: dec]
	C: [CPU:inc, Heap: inc, RAM: inc, NW: inc, Disk: inc]
```

#### First we create an impact matrix for the cluster:
* node A has only decrease across each dim
* B has 1 inc and 1 dec action, 
* C has 1 inc action for each dim

```ruby
	       A 	B 	      C
cpu 	  i=0,d=1     i=1,d=1     i=1,d=0
heap 	  i=0,d=1     i=1,d=1     i=1,d=0
ram       i=0,d=1     i=1,d=1     i=1,d=0
nw        i=0,d=1     i=1,d=1     i=1,d=0
disk      i=0,d=1     i=1,d=1     i=1,d=0
```

#### Now we run the actions through this matrix again.
> If an action is increasing pressure on a dim on a node, it is only allowed 
> if there is no other action trying to dec that dim on the node. i.e. d=0 for that node,dim.

#### When an action gets purged in this second pass, we remove it from the matrix by reducing the node,dim counters.

In above e.g. we process `MoveShard(A->B)` - 
* It cannot go through because `d=1` for some dims on node B
* We purge `MoveShard(A->B)`
* Counters on A and B are updated. All counters in A become `i=0, d=0`. B becomes `i=0,d=1`

Now we process `MoveShard(B->C)` - 
* It increases pressure on C but there is no one decreasing pressure (`d=0` for all dims on C), so it can go through 
* On B it is dec pressure anyway so it has a clear pass.

Hence `MoveShard(B->C)` gets picked.

This will also resolve cycles like A->B, B->C, C->A by picking either one or two of the actions (depending on order of 2nd pass).

---

It is okay if we handle multi-node actions in a separate PR later. You may also use different/better algorithms for polarizing impact vectors. Just wanted to check if the current `Collator::polarize()` and `ActionGrouper::groupByInstanceId()` is the right approach for future use cases, like the moveShard, splitShard etc..

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
```diff
@@ -0,0 +1,148 @@
...
+
+  @NonNull
+  private List<Action> getAllProposedActions() {
+    final List<Action> proposedActions = new ArrayList<>();
+    if (deciders != null) {
+      for (final Decider decider : deciders) {
+        List<Decision> decisions = decider.getFlowUnits();
+        if (decisions != null) {
+          decisions.forEach(decision -> {
+            if (decision.getActions() != null) {
```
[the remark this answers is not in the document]
ktkrg: will change in the next rev

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
ktkrg: Makes sense, I assumed for the first iteration we would not have such actions and had not thought of polarizing at a dimension level. I will change it.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/Collator.java
ktkrg: Thanks for this sage advice, I had an unnecessarily complex system and I was able to rip it out until I got to something similar to what you've proposed here :) I've updated the PR to reflect these changes.

yojs:
Do you think we can add integ tests for this ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/ImpactAssessment.java
```diff
@@ -0,0 +1,154 @@
...
+   * Alignment is checked when reassessing the actions where all impacts are replayed.
+   *
+   * @param actionName The name of the action.
+   * @param impactVector The impact vector which gives a pressure heading for various resources
+   *                     impacted by taking this action.
+   * @return true if the impact vector aligns with the overall impact, false otherwise.
+   */
+  public boolean checkAlignmentAcrossDimensions(@NonNull final String actionName,
+      @NonNull final ImpactVector impactVector) {
+    boolean isAligned = true;
```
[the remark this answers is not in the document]
ktkrg: Done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/ImpactAssessment.java
```diff
@@ -0,0 +1,154 @@
...
+
+    // If this is an action that increases pressure along some dimension for this node, and the
+    // overall assessment says there are actions that decrease pressure along those same
+    // dimensions, then this action is not aligned with the other proposed actions where the
+    // deciders are trying to reduce pressure for those dimensions.
+
+    final Map<Dimension, Impact> impactMap = impactVector.getImpact();
+    for (final Map.Entry<Dimension, Impact> entry : impactMap.entrySet()) {
+      final Impact impactOnDimension = entry.getValue();
```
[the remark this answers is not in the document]
ktkrg: We can't use `forEach` here because we will be updating a local variable inside the body of the lambda and it needs the `isAligned` value to be effectively final. Since local variables are not effectively final, we'll need to declare an array or a list of size=1 to hold the `isAligned` value.

I chose the verbose way to avoid some confusion there.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/ImpactAssessment.java
```diff
@@ -0,0 +1,154 @@
...
+      if (isAligned && impactOnDimension.equals(Impact.INCREASES_PRESSURE)) {
+        List<String> pressureDecreasingActions = perDimensionPressureDecreasingActions
+            .getOrDefault(entry.getKey(), Collections.emptyList());
+        isAligned = pressureDecreasingActions.isEmpty();
```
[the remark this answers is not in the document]
ktkrg: Done.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/ImpactAssessment.java
```diff
@@ -0,0 +1,154 @@
...
+      }
+    }
+
+    return isAligned;
+  }
+
+  private void addActionToMap(@NonNull final Map<Dimension, List<String>> map,
+      @NonNull final String actionName, @NonNull final Dimension dimension) {
+    map.computeIfAbsent(dimension,
+        dim -> new ArrayList<>()).add(actionName);
```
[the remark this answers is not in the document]
ktkrg: Done.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/collator/CollatorTest.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  public ImpactVector buildShardMoveInImpactVector() {
+    final ImpactVector impactVector = new ImpactVector();
+    impactVector.increasesPressure(Dimension.CPU);
+    impactVector.increasesPressure(Dimension.HEAP);
+
+    return impactVector;
+  }
+}
```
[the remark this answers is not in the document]
ktkrg: Done.

Issue #339, opened 2020-08-05
Populate decider configs as a part of the node startup. 


aditjind:
Take workload type / cache priority order as an input from the customer so as to default decider actions based on these inputs. Also have a generic decider config which would be populated whenever any decider node is added to the RCA graph.

PR #340, opened 2020-08-05
Populate default workload type and cache priority for the decider to base default actions

aditjind:
*Issue #:* #339 

*Description of changes:* These changes take the workload type and cache priorities as an input from the customer to base the actions taken on the cluster. Default values are hardcoded in the code in case the read from config is not possible due to any circumstances. Values present in the config file are given priority over any such value. 

*Tests:* UTs added. 


pa_config/rca.conf
```diff
@@ -87,8 +87,22 @@
   ],
 
   "decider-config-settings": {
+    // Decreasing order of priority for the type of workload we can expect on the cluster.
+    "workload-type": {
+      "priority-order":[
+        "ingest",
+        "search"
+      ]
+    }
```
sidheart, line 96: I'm pretty sure you need commas between these JSON objects

pa_config/rca_idle_master.conf
```diff
@@ -98,8 +98,22 @@
   ],
 
   "decider-config-settings": {
+    // Decreasing order of priority for the type of workload we can expect on the cluster.
+    "workload-type": {
+      "priority-order":[
+        "ingest",
+        "search"
+      ]
```
sidheart, line 106: Commas, see above

pa_config/rca_master.conf
```diff
@@ -98,8 +98,22 @@
   ],
 
   "decider-config-settings": {
+    // Decreasing order of priority for the type of workload we can expect on the cluster.
+    "workload-type": {
```
sidheart, line 102: Add commas between objects

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decider.java
```diff
@@ -90,4 +95,31 @@ public void handleNodeMuted() {
...
+   * read threshold values from rca.conf
+   *
+   * @param conf RcaConf object
+   */
+  @Override
+  public void readRcaConf(RcaConf conf) {
+    configObj = conf.getDeciderConfig();
+  }
+
+  public double getFieldDataCacheUpperBound() {
```
sidheart, line 109: Needs null checks

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/DeciderConfig.java
```diff
@@ -0,0 +1,80 @@
...
+    public static final String PRIORITY_ORDER_CONFIG_NAME = "priority-order";
+    public static final String FIELD_DATA_CACHE_UPPER_BOUND = "field-data-cache-upper-bound";
+    public static final String SHARD_REQUEST_CACHE_UPPER_BOUND = "shard-request-cache-upper-bound";
+    private static final double DEFAULT_FIELD_DATA_CACHE_UPPER_BOUND = 0.4;
+    private static final double DEFAULT_SHARD_REQUEST_CACHE_UPPER_BOUND = 0.05;
+    // Defaults based on prioritising Stability over performance.
+    private static final List<String> DEFAULT_WORKLOAD_PRIORITY = Arrays.asList("ingest", "search");
+    private static final List<String> DEFAULT_CACHE_PRIORITY = Arrays.asList("read-heavy", "write-heavy");
+
+    public Double fieldDataCacheUpperBound;
```
sidheart, line 37: It seems like these are written and then never read. Are we doing this on purpose? Will this only be configurable in the future?

sidheart: Nvm, I see what's happening here. Can you follow the Java standard practice and make these private and add getters for them?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decider.java
sidheart: (on configObj)

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decider.java
```diff
@@ -90,4 +95,31 @@ public void handleNodeMuted() {
...
+    return configObj.shardRequestCacheUpperBound;
+  }
+
+  public List<String> getWorkLoadPriority() {
+    return configObj.workloadPriorityOrder;
+  }
+
+  public List<String> getCachePriority() {
+    return configObj.cachePriorityOrder;
+  }
```
sidheart, line 123: Left a comment on DeciderConfig as well. Please use private members + getters for this

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
```diff
@@ -69,8 +70,8 @@ public CacheHealthDecider(
             .put(ResourceEnum.FIELD_DATA_CACHE, fieldDataCacheClusterRca)
             .build();
 
-    this.fieldDataCacheSizeUpperBound = CacheDeciderConfig.DEFAULT_FIELD_DATA_CACHE_UPPER_BOUND;
-    this.shardRequestCacheSizeUpperBound = CacheDeciderConfig.DEFAULT_SHARD_REQUEST_CACHE_UPPER_BOUND;
+    this.fieldDataCacheSizeUpperBound = getDefaultShardRequestCacheUpperBound();
+    this.shardRequestCacheSizeUpperBound = getDefaultFieldDataCacheUpperBound();
```
sidheart, line 74: Add a comment on why we're always going with the default here instead of allowing overrides.

sidheart: These fields also make certain calls in this class super misleading. e.g. line 167 on this file has a call to getFieldDataCacheUpperBound(), but that calls Decider#getFieldDataCacheUpperBound which reads the value from a configObj and not this field right? What's the purpose of these fields?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/DeciderConfig.java
aditjind: Have added for the ones which are not being used currently as well.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decider.java
aditjind: Done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
aditjind: I am not sure on the purpose of these fields. I have removed these fields as the the default value would be added in case the config is not read from the file.

pa_config/rca.conf
```diff
@@ -86,11 +86,21 @@
...
+  // Default Values are added in the config specific file as well
+  // Modify them as well whenever changing the default values (DeciderConfig.java)
   "decider-config-settings": {
+    // Decreasing order of priority for the type of workload we can expect on the cluster.
+    "workload-type": {
+      "priority-order": ["ingest", "search"]
+    },
+    // Decreasing order of priority for the type of cache which is expected to be consumed more.
+    "cache-type": {
+      "priority-order": ["read-heavy", "write-heavy"]
```
rguo-aws, line 98: a high level question is how do we plan to use this cache priority in decider. I understand the purpose of having cache is to accelerate reading. So what should decider do if write heavy is specified here?  
And if customer wants to prioritized fielddata cache over shard request cache, how should they set it in config  ?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/RcaConfTest.java
```diff
@@ -63,20 +64,14 @@ public void testReadRcaConfig() {
...
-    Assert.assertNotNull(shardRequestCacheUpperBound);
-    Assert.assertEquals(CacheDeciderConfig.DEFAULT_SHARD_REQUEST_CACHE_UPPER_BOUND, shardRequestCacheUpperBound, 0.01);
+    DeciderConfig configObj = new DeciderConfig(rcaConf);
+    Assert.assertNotNull(configObj.getCachePriorityOrder());
+    Assert.assertNotNull(configObj.getWorkloadPriorityOrder());
+    Assert.assertNotNull(configObj.getFieldDataCacheUpperBound());
+    Assert.assertNotNull(configObj.getShardRequestCacheUpperBound());
+    Assert.assertEquals(Arrays.asList("test-read-heavy", "test-write-heavy"), configObj.getCachePriorityOrder());
+    Assert.assertEquals(10.4, configObj.getFieldDataCacheUpperBound(), 0.01);
+    Assert.assertEquals(10.05, configObj.getShardRequestCacheUpperBound(), 0.01);
```
sruti1312, line 74: 0.4 and 0.05? Any reason why we are using 10.4 and 10.05?

pa_config/rca_master.conf
```diff
@@ -97,11 +97,21 @@
...
+    "cache-type": {
+      "priority-order": ["read-heavy", "write-heavy"]
+    },
     // cache decider - Needs to be updated as per the performance test results
-    "cache-decider-config": {
-      "field-data-cache-upper-bound" : 0.4,
-      "shard-request-cache-upper-bound" : 0.05
+    "cache-bounds": {
+      "field-data-cache-upper-bound" : 10.4,
+      "shard-request-cache-upper-bound" : 10.05
```
sruti1312, line 114: What are your thoughts on having one more level. Like all cache related config bounded together. Pro might be it becomes more manageable.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/RcaConfTest.java
aditjind: I have updated the values in the tmp config files so as to assert the reading from the config is working as expected.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/DeciderConfig.java
```diff
@@ -0,0 +1,120 @@
...
+
+    public static String getWorkloadConfigName() {
+        return WORKLOAD_CONFIG_NAME;
+    }
+
+    public static String getPriorityOrderConfigName() {
+        return PRIORITY_ORDER_CONFIG_NAME;
+    }
+
+    public static String getFieldDataCacheUpperBoundName() {
```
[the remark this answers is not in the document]
aditjind: I have created getters for all the private fields. We don't need them now but maybe if want to re-use the variables in some other module, it would be nice to keep it here.

pa_config/rca_master.conf
aditjind: We don't anticipate any more configs just now. So I will be more tilted to keep it just like this. I agree if too many configs are available then it would be nice to club similar of them together.

pa_config/rca_idle_master.conf
```diff
@@ -97,11 +97,21 @@
...
+    // Decreasing order of priority for the type of cache which is expected to be consumed more.
+    "cache-type": {
+      "priority-order": ["read-heavy", "write-heavy"]
+    },
     // cache decider - Needs to be updated as per the performance test results
-    "cache-decider-config": {
-      "field-data-cache-upper-bound" : 0.4,
-      "shard-request-cache-upper-bound" : 0.05
+    "cache-bounds": {
+      "field-data-cache-upper-bound" : 10.4,
```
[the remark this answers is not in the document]
aditjind: Good catch. I had modified the values for the UT config and by mistake forgot to modify the defaults here. Updated.

pa_config/rca.conf
aditjind: The assumption here is customer doesn't know in detail about the field data cache and the shard request cache. So customer is going to provide the inputs in terms of there workload priority and cache type priority. Our deciders then in turn would change these variables into the prioritising Field data cache over shard request cache.

vigyasharma: Caches are different from queues and don't easily map to workload types. From what i know, all caches help with some form of read traffic in ES. Indexing has a separate indexing buffer.
For caches, customers will have to call out the priority wrt. specific cache names - `field-data-cache, shard-request-cache` etc. A read/write heavy cluster does not provide data on relative cache priorities. 
These caches are public documentation so customers would know about it, but it would still be an advanced setting.

pa_config/rca_master.conf
```diff
@@ -102,9 +102,19 @@
...
+    "workload-type": {
+      "priority-order": ["ingest", "search"]
+    },
+    // Decreasing order of priority for the type of cache which is expected to be consumed more.
+    "cache-type": {
+      "priority-order": ["read-heavy", "write-heavy"]
+    },
     // cache decider - Needs to be updated as per the performance test results
-    "cache-decider-config": {
+    "cache-bounds": {
```
vigyasharma, line 117: We should move default values to code and pick overrides from `rca.conf`. This can be a separate PR/task.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/DeciderConfig.java
```diff
@@ -0,0 +1,120 @@
...
+public class DeciderConfig {
+
+    private static final String CACHE_BOUNDS_CONFIG_NAME = "cache-bounds";
+    private static final String CACHE_CONFIG_NAME = "cache-type";
+    private static final String WORKLOAD_CONFIG_NAME = "workload-type";
+    private static final String PRIORITY_ORDER_CONFIG_NAME = "priority-order";
+    private static final String FIELD_DATA_CACHE_UPPER_BOUND_NAME = "field-data-cache-upper-bound";
+    private static final String SHARD_REQUEST_CACHE_UPPER_BOUND_NAME = "shard-request-cache-upper-bound";
+    private static final double DEFAULT_FIELD_DATA_CACHE_UPPER_BOUND = 0.4;
+    private static final double DEFAULT_SHARD_REQUEST_CACHE_UPPER_BOUND = 0.05;
```
vigyasharma, line 32: We can split cache bounds and queue bounds into their respective config objects. They will still get read from the decider base class so all deciders have it. But this will help organize the overall structure. (`QueueConfig` will get passed to `ModifyQueueAction` etc. )
We can do this as a separate PR.

pa_config/rca_idle_master.conf
```diff
@@ -102,9 +102,19 @@
...
   "muted-actions": [],
 
+  // Default Values are added in the config specific file as well
+  // Modify them as well whenever changing the default values (DeciderConfig.java)
   "decider-config-settings": {
+    // Decreasing order of priority for the type of workload we can expect on the cluster.
+    "workload-type": {
+      "priority-order": ["ingest", "search"]
+    },
+    // Decreasing order of priority for the type of cache which is expected to be consumed more.
```
vigyasharma, line 112: Minor: Consider rewording to make the ordering more clear. It should be clear that the list goes from most used to least used cache. (Same for workload priority).

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Decider.java
```diff
@@ -91,4 +101,31 @@ public void handleNodeMuted() {
...
+
+  public double getFieldDataCacheUpperBound() {
+    return configObj != null ? configObj.getFieldDataCacheUpperBound() : getDefaultFieldDataCacheUpperBound();
+  }
+
+  public double getShardRequestCacheUpperBound() {
+    return configObj != null ? configObj.getShardRequestCacheUpperBound() : getDefaultShardRequestCacheUpperBound();
+  }
+
+  public List<String> getWorkLoadPriority() {
```
vigyasharma, line 123: Good decision keeping them in base class. These will be required by multiple deciders.

pa_config/rca_master.conf
aditjind: This is how it is happening currently. We read the overrides only from the `rca.conf` and default values are hard coded. Currently default and overridden values are the same.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/DeciderConfig.java
aditjind: So as suggested cache bounds / queue bounds etc. should be read as a single object from the config file? This can be done when the queue related configs are added in the config file.

pa_config/rca_idle_master.conf
aditjind: done

pa_config/rca.conf
aditjind: So do you suggest we should remove the read-heavy / write-heavy input altogether? If we don't want to use this for cache then the other things we needed it for can be derived from workload (search/ingest) type?

vigyasharma: My suggestion is to directly input cache priorities, i.e. relevant caches for the customer from most imp to least imp (or vice versa). You can think of what the config key should be, but values will contain actual cache names. 

Something like - 
```ruby
"cache-priority": ["fielddata-cache", "shard-request-cache", "query-cache", "bitset-filter-cache"]
```

We cannot derive this from workload (search/ingest) types.

aditjind: Right! I have updated the default values to specific caches. The output we wanted to derive from read-heavy / write-heavy can be taken from search / ingest as well but we do need the explicit cache priority.

pa_config/rca.conf
```diff
@@ -91,9 +91,21 @@
   "muted-deciders": [],
   "muted-actions": [],
 
+  // Default Values are added in the config specific file as well
+  // Modify them as well whenever changing the default values (DeciderConfig.java)
   "decider-config-settings": {
+    // Decreasing order of priority for the type of workload we can expect on the cluster.
+    // Priority order in the list goes from most expected to the lease expected workload type.
```
sruti1312, line 98: nit: least*

pa_config/rca.conf
```diff
@@ -91,9 +91,21 @@
...
+    "workload-type": {
+      "priority-order": ["ingest", "search"]
+    },
+    // Decreasing order of priority for the type of cache which is expected to be consumed more.
+    // Priority order in the list goes from most used to the lease used cache type.
```
sruti1312, line 103: same*

Issue #347, opened 2020-08-06
Node Summary Serialization Issue for Cache RCAs


khushbr:
Serialization using ‘buildFlowUnitMessage()’ in ResourceFlowUnit failing. 

```
[2020-08-05T15:47:43,681][ERROR][c.a.o.e.p.r.f.a.f.ResourceFlowUnit] RCA: casting to wrong summary type when de-serializing this flowunit
[2020-08-05T15:47:43,681][ERROR][c.a.o.e.p.r.f.a.f.ResourceFlowUnit] RCA: casting to wrong summary type when de-serializing this flowunit

```

PR #348, opened 2020-08-06
Fixing the summary serialization issue for cache RCAs

khushbr:
*Issue #:* https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/issues/347

*Description of changes:* Serialization failing for Cache RCAs because of `null HotNodeSummary` when Resource is Healthy.
Updated the default initialization and removed the Null assignment for `HotNodeSummary`


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/FieldDataCacheRca.java
```diff
@@ -112,21 +112,19 @@ public void setClock(Clock clock) {
         cacheEvictionCollector.collect(currTimestamp);
         if (counter >= rcaPeriod) {
             ResourceContext context;
-            HotNodeSummary nodeSummary;
-
             InstanceDetails instanceDetails = getInstanceDetails();
+            HotNodeSummary nodeSummary = new HotNodeSummary(instanceDetails.getInstanceId(), instanceDetails.getInstanceIp());
```
vigyasharma, line 116: So is the problem here a `null HotNodeSummary` ? Can we add UTs for a case where only few (or no) nodes have an issue?

PR #351, opened 2020-08-07
Fix bug in publisher to support cool off period on a per node basis

rguo-aws:
*Issue #:*
#352 

*Description of changes:*
The current implementation of cool off action map only take action.name() as the hashkey and map it to the execution time. This is incorrect because if let's say we have two actions of the same type but has different target (impacted) nodes. One action will be published while the other will be blocked. This is not the expected behavior as both actions should be published if their impacted nodes are not overlapped. We should use action.name() + nodekey as the hashkey instead.

In this PR, I moved the cool off action map into a separate class namely CoolOffDetector and extend the hashmap to support hashkey as action.name() + nodekey.

*Tests:*
n/a


yojs:
Can we add an issue for this PR ?

PR #356, opened 2020-08-07
Expose queue rejection time period setting in rca.conf

rguo-aws:
*Issue #:*
#357 

*Description of changes:*
Expose queue rejection time period setting in rca.conf

*Tests:*
tested on docker


pa_config/rca.conf
```diff
@@ -49,6 +49,10 @@
       "promotion-rate-mb-per-second" : 500,
       "young-gen-gc-time-ms-per-second" : 400
     },
+    //queue rejection rca
```
yojs, line 52: The comment is the key name itself. It may be removed.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/QueueRejectionRcaConfig.java
```diff
@@ -0,0 +1,40 @@
...
+      rejectionTimePeriodInSeconds = DEFAULT_REJECTION_TIME_PERIOD_IN_SECONDS;
+    }
+  }
+
+  public int getRejectionTimePeriodInSeconds() {
+    return rejectionTimePeriodInSeconds;
+  }
+
+  public static class RCA_CONF_KEY_CONSTANTS {
+    public static final String REJECTION_TIME_PERIOD_IN_SECONDS = "rejection-time-period-in-seconds";
```
yojs, line 38: Is there a reason why it should be in a static inner class of its own ?

pa_config/rca.conf
rguo-aws: removed

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/configs/QueueRejectionRcaConfig.java
rguo-aws: the reason is because we might want to add more settings to this RCA in the future so it would be better to keep all consts under the same inner class.

PR #371, opened 2020-08-13
Refactor ModifyCacheCapacityAction to follow builder pattern

sruti1312:
*Description of changes:*
Refactor ModifyCacheCapacityAction to follow builder pattern. 
This allows modifying parameters by deciders

*Tests:*
unit test


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -34,49 +33,44 @@
...
       final ResourceEnum cacheType,
-      final NodeConfigCache nodeConfigCache,
-      final double cacheSizeUpperBound,
-      final boolean increase,
-      final AppContext appContext) {
-    // TODO: Add lower bound for caches
-    // TODO: Address cache scaling down  when JVM decider is available
-
+      final AppContext appContext,
+      final long desiredCacheMaxSizeInBytes,
```
rguo-aws, line 53: long primitive itself is immutable. We don't need to explicitly specify it as final

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -133,53 +122,120 @@ public String toString() {
     return summary();
   }
 
-  public void setStepSizeForCache(final long stepSizeInBytes) {
```
rguo-aws, line 136: same as above

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -133,53 +122,120 @@ public String toString() {
...
 
-  private boolean validateAndSetConfigValues() {
-    // This is intentionally not made static because different nodes can
-    // have different bounds based on instance types
+  public static final class Builder {
+    public static final long DEFAULT_COOL_OFF_PERIOD_IN_MILLIS = 300 * 1_000;
+    public static final boolean DEFAULT_IS_INCREASE = true;
+    public static final boolean DEFAULT_CAN_UPDATE = true;
+    // TODO: Update the step size to also include percentage of heap size along with absolute value
+    public static final ImmutableMap<ResourceEnum, Long> DEFAULT_STEP_SIZE_IN_BYTES =
```
rguo-aws, line 142: Do we still need a static map here ? let's use switch for now and later on we will read the default value directly from RcaConf object

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -133,53 +122,120 @@ public String toString() {
...
+        new ImmutableMap.Builder<ResourceEnum, Long>()
+            .put(
+                ResourceEnum.FIELD_DATA_CACHE,
+                512 * 1_024L * 1_024L) // Field data cache having step size of 512MB
```
rguo-aws, line 146: Let's define a static const/ or static function somewhere to convert byte to KB/MB

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -133,53 +122,120 @@ public String toString() {
...
+
+    private Long currentCacheMaxSizeInBytes;
+    private Long desiredCacheMaxSizeInBytes;
+    private Long heapMaxSizeInBytes;
+
+    public Builder(
+        final NodeKey esNode,
+        final ResourceEnum cacheType,
+        final AppContext appContext,
+        final double upperBoundThreshold) {
```
rguo-aws, line 170: we should read upper bound / lower bound from RcaConf. Let's use a switch to assign it to default for now

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
sruti1312: The intent of adding final to these values to make them immutable. explicitly declare it as final so we won’t try to make any changes to that variable within the constructor. 

Based on our discussion. Removing final with setter's and retaining them in the constructors.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
sruti1312: The intent of adding final to these values to make them immutable. explicitly declare it as final so we won’t try to make any changes to that variable within the constructor. 

Based on our discussion. Removing final with setter's and retaining them in the constructors.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
sruti1312: Presently the cache decider is reading from rcaConf. It parses and sends the required fields to actions. It makes sense to retain this. Aditya has a PR open that makes changes to deciderConfig structure. After which there will be a modification in this.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
sruti1312: Changed to a switch.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
sruti1312: Added util functions

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -133,53 +122,127 @@ public String toString() {
...
+        case FIELD_DATA_CACHE:
+          // Field data cache having step size of 512MB
+          this.stepSizeInBytes = (long) CacheUtil.getMBSizeInBytes(512);
+          break;
+        case SHARD_REQUEST_CACHE:
+          // Shard request cache step size of 512KB
+          this.stepSizeInBytes = (long) CacheUtil.getKBSizeInBytes(512);
+          break;
+        default:
+          assert false : "unrecognized cache type: " + cacheType.name();
```
khushbr, line 192: Add a metric here?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -133,53 +122,127 @@ public String toString() {
...
-  }
+    public Builder upperBoundThreshold(double upperBoundThreshold) {
+      this.upperBoundThreshold = upperBoundThreshold;
+      return this;
+    }
 
-  private void setDesiredCacheMaxSize(final long desiredCacheMaxSize) {
-    this.desiredCacheMaxSizeInBytes = Math.min(desiredCacheMaxSize, cacheUpperBoundInBytes);
+    public ModifyCacheMaxSizeAction build() {
+      // fail to read max size from node config cache
```
khushbr, line 222: nit: Update to `In case of failure to read max size from node config cache`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -133,53 +122,127 @@ public String toString() {
...
+      // return an empty non-actionable action object
+      if (currentCacheMaxSizeInBytes == null || heapMaxSizeInBytes == null) {
+        LOG.error(
+            "Action: Fail to read cache max size or heap max size from node config cache. Return an non-actionable action");
+        return new ModifyCacheMaxSizeAction(
+            esNode, cacheType, appContext, -1, -1, coolOffPeriodInMillis, false);
```
khushbr, line 228: Let's use `Null` instead of `-1`, ES interprets `-1` cache size as unbounded and I understand we are setting the `canUpdate` as `False` but let's update this to avoid giving the wrong impression.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -133,53 +122,127 @@ public String toString() {
...
   }
 
   public ResourceEnum getCacheType() {
     return cacheType;
   }
 
-  private boolean validateAndSetConfigValues() {
-    // This is intentionally not made static because different nodes can
-    // have different bounds based on instance types
+  public static final class Builder {
```
khushbr, line 137: Can we make the class as `private` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
rguo-aws: @khushbr we should not run into this default statement unless there is a bug in decider. This assert help us to find bug in early development phase and adding a metric is a overkill as this is not runtime error and we are not expected to run into it

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -34,49 +33,44 @@
...
+    this.coolOffPeriodInMillis = coolOffPeriodInMillis;
+    this.canUpdate = canUpdate;
+  }
 
-    if (validateAndSetConfigValues()) {
-      long desiredCapacity =
-              increase ? currentCacheMaxSizeInBytes + getStepSize(cacheType) : currentCacheMaxSizeInBytes;
-      setDesiredCacheMaxSize(desiredCapacity);
-    }
+  public static Builder newBuilder(
```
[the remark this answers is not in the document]
rguo-aws: we try to follow the same naming convention that protobuf uses. They named the static method as newBuilder

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -133,53 +122,127 @@ public String toString() {
...
+      setDefaultStepSize(cacheType);
+    }
+
+    private void setDefaultStepSize(ResourceEnum cacheType) {
+      // TODO: Move configuration values to rca.conf
+      // TODO: Update the step size to also include percentage of heap size along with absolute value
+      switch (cacheType) {
+        case FIELD_DATA_CACHE:
+          // Field data cache having step size of 512MB
+          this.stepSizeInBytes = (long) CacheUtil.getMBSizeInBytes(512);
```
rguo-aws, line 185: this stepsize is a predefined static value which never get updated in runtime. Instead of creating a util function, can we simply add 
this.stepSizeInBytes = (long) 512*MB_to_BYTES
And define MB_to_BYTES as a static variable somewhere ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
rguo-aws: @khushbr we did this intentionally as per Vigya's comment. We generate a default non-actionable Action object here to avoid handling null pointer in caller

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -133,53 +122,127 @@ public String toString() {
...
+                      appContext.getNodeConfigCache(), esNode, cacheType);
+      this.heapMaxSizeInBytes =
+              NodeConfigCacheReaderUtil.readHeapMaxSizeInBytes(appContext.getNodeConfigCache(), esNode);
+      this.desiredCacheMaxSizeInBytes = null;
+      setDefaultStepSize(cacheType);
+    }
+
+    private void setDefaultStepSize(ResourceEnum cacheType) {
+      // TODO: Move configuration values to rca.conf
+      // TODO: Update the step size to also include percentage of heap size along with absolute value
```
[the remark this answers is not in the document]
rguo-aws: Agree. We should adjust step size based on heap size. We can either adjust it in Action itself or we can do it in decider.
Any thought ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -34,49 +33,44 @@
...
  * cache and shard request cache.
  */
-
-// TODO: Split the cache action into separate actions for different caches.
-
 public class ModifyCacheMaxSizeAction extends SuppressibleAction {
   private static final Logger LOG = LogManager.getLogger(ModifyCacheMaxSizeAction.class);
-  public static final String NAME = "modifyCacheCapacity";
-  public static final long COOL_OFF_PERIOD_IN_MILLIS = 300 * 1_000;
+  public static final String NAME = "ModifyCacheCapacity";
```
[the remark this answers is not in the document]
sruti1312: Right. Modified

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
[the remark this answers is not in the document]
sruti1312: TODO: Include percentage with (and) absolute value. I think the default can be adjusted in the action itself and provide decider the functionality to override this value. We can have an offline discussion about this.

Right. Will address this in subsequent PR.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
sruti1312: Modified

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
sruti1312: Yes we can make this private. Modified

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
sruti1312: Changed to define a static variable instead of a function

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
sidheart: Please throw an IllegalArgumentException() here and annotate the method. I could easily mistakenly call this with ResourceEnum.OLD_GEN and crash the program.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/CacheUtil.java
```diff
@@ -29,6 +29,9 @@
 public class CacheUtil {
     private static final Logger LOG = LogManager.getLogger(CacheUtil.class);
 
+    public static final long MB_TO_BYTES = 1024 * 1024;
```
sidheart, line 32: supernit: KB_TO_BYTES * 1024

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeActionTest.java
```diff
@@ -74,23 +74,46 @@ public void setUp() {
   public void testIncreaseCapacity() {
     NodeKey node1 =
         new NodeKey(new InstanceDetails.Id("node-1"), new InstanceDetails.Ip("1.2.3.4"));
-    ModifyCacheMaxSizeAction modifyCacheSizeAction =
-        new ModifyCacheMaxSizeAction(
+    ModifyCacheMaxSizeAction.Builder builder =
```
sidheart, line 77: This class is missing 2 important tests. When the Builder is unable to determine currentCacheMaxSizeInBytes or heapMaxSizeInBytes, you need to verify that build() returns a non-actionable action.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/cache/CacheUtil.java
sruti1312: Modified

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeActionTest.java
sruti1312: Added tests around this

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -133,53 +123,127 @@ public String toString() {
...
+      this.currentCacheMaxSizeInBytes =
+              NodeConfigCacheReaderUtil.readCacheMaxSizeInBytes(
+                      appContext.getNodeConfigCache(), esNode, cacheType);
+      this.heapMaxSizeInBytes =
+              NodeConfigCacheReaderUtil.readHeapMaxSizeInBytes(appContext.getNodeConfigCache(), esNode);
+      this.desiredCacheMaxSizeInBytes = null;
+      setDefaultStepSize(cacheType);
+    }
+
+    private void setDefaultStepSize(ResourceEnum cacheType) {
```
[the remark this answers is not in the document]
sruti1312: setDefaultStepSize is a private method that only the constructor uses. 
Similarly Builder is also private. We call builder from a static method newBuilder that returns this object. We dont need to return it explicitly.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
sruti1312: Right. throwing an IllegalArgumentException

PR #383, opened 2020-08-17
Improve RCA Testing Workflow

sidheart:
*Fixes #:* 

*Description of changes:*
- Annotating a test with @category(SlowTests.class) will cause the
default :test task to skip the test. These tests can be run as described
in the bullet below.

- The :test gradle task will now only run a subset of fast tests by
default. To run the RCA Integration Tests (e.g. RcaItQueueTuning) you
must provide the -PrunITs=true flag to gradle. To run tests in the newly
added SlowTests category, you must pass -PrunSlowTests=run to gradle. So
the command to run all tests is ./gradlew test -PrunITs=true
-PrunSlowTests=true.

- Workflows have been modified so that the PerformanceAnalyzerIT
Integration Test suite (the test suite that hits our APIs) runs on push
to the performance-analyzer-rca master or dev branches.

- Workflows have been modified so that PRs undergo minimal testing,
while pushes to master or dev undergo the full suite of testing. If
there is an error on master, we should revert first and ask questions
later. In the future, we can automate this reversion flow.

*Tests:* N/A


sidheart:
I've broken the relevant pieces of this commit down into smaller PRs

PR #386, opened 2020-08-19
Create a PersistableAction base class to persist actions into database

rguo-aws:
*Fixes #:*

*Description of changes:*
This PR allows us to persist Action into "Action" table in SQL. So we create a base class PersistableAction to store SQL schema/tuple and the method to persist tuple into the corresponding table and extend Modify*Action from the PersistableAction. 

Since a Decision made by decider can have multiple actions so we want to create a PersistEvent to keep the one-to-many relation into DB as well. So we create a separate table namely "PersistorEvent" to store decisions emit by collator and link actions in Action table to this element.

*Tests:*
UTs is still in progress  

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
@@ -20,6 +20,7 @@ import static com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionma
 import static com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.ImpactVector.Dimension.NETWORK;
 
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.AppContext;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.PersistableAction.SQL_SCHEMA_CONSTANTS.ActionField;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.ResourceEnum;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.NodeKey;
 
@@ -30,7 +31,7 @@ import java.util.Map;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-public class ModifyQueueCapacityAction extends SuppressibleAction {
+public class ModifyQueueCapacityAction extends PersistableAction {
 
   private static final Logger LOG = LogManager.getLogger(ModifyQueueCapacityAction.class);
   public static final String NAME = "ModifyQueueCapacity";
@@ -56,6 +57,7 @@ public class ModifyQueueCapacityAction extends SuppressibleAction {
     this.lowerBound = lowerBound;
     this.upperBound = upperBound;
     this.canUpdate = canUpdate;
+    persistValues();
   }
 
   public static Builder newBuilder(NodeKey esNode, ResourceEnum threadPool, AppContext appContext) {
@@ -117,6 +119,17 @@ public class ModifyQueueCapacityAction extends SuppressibleAction {
 
   public ResourceEnum getThreadPool() {
     return threadPool;
+  }
+
+  //persist values into SQL tuple
+  private void persistValues() {
+    setRowValue(ActionField.RESOURCE_TYPE_FIELD.getField(), threadPool.getNumber());
+    setRowValue(ActionField.NODE_ID_FIELD.getField(), esNode.getNodeId().toString());
+    setRowValue(ActionField.NODE_IP_FIELD.getField(), esNode.getHostAddress().toString());
+    setRowValue(ActionField.ACTIONABLE_FIELD.getField(), isActionable());
+    setRowValue(ActionField.COOL_OFF_PERIOD_FIELD.getField(), coolOffPeriodInMillis);
+    setRowValue(ActionField.CURRENT_VALUE_FIELD.getField(), currentCapacity);
+    setRowValue(ActionField.DESIRED_VALUE_FIELD.getField(), desiredCapacity);
   }
 
   public static final class Builder {
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/PersistableAction.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/PersistableAction.java
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions;
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.AppContext;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.Consts.Table;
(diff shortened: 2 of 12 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -56,6 +57,7 @@ private ModifyQueueCapacityAction(NodeKey esNode, ResourceEnum threadPool, AppCo
     this.lowerBound = lowerBound;
     this.upperBound = upperBound;
     this.canUpdate = canUpdate;
+    persistValues();
```
vigyasharma, line 60: Why do we want to persist values as soon as an action is created? Shouldn't we just persist them when an action is published, why would we log in db the actions that were created but discarded or pruned?

The `Persistable` interface which actions implement can have a `persistToDb()` fn that a `PublisherEventsPersistor implements ActionListener` calls. (can rename to better names)

Or better still, we can have the `ActionListener` hold all the logic for persisting to sql, and just call `Action` methods like `name(), impactedNodes(), summary(), coolOffPeriodInMillis(), isActionable()` to get values and persist them.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -119,6 +121,17 @@ public ResourceEnum getThreadPool() {
     return threadPool;
   }
 
+  //persist values into SQL tuple
+  private void persistValues() {
+    setRowValue(ActionField.RESOURCE_TYPE_FIELD.getField(), threadPool.getNumber());
+    setRowValue(ActionField.NODE_ID_FIELD.getField(), esNode.getNodeId().toString());
+    setRowValue(ActionField.NODE_IP_FIELD.getField(), esNode.getHostAddress().toString());
```
vigyasharma, line 128: A single action can impact multiple nodes in future - like `moveShard` actions. These fields should allow for multiple node values

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -119,6 +121,17 @@ public ResourceEnum getThreadPool() {
...
+    setRowValue(ActionField.ACTIONABLE_FIELD.getField(), isActionable());
+    setRowValue(ActionField.COOL_OFF_PERIOD_FIELD.getField(), coolOffPeriodInMillis);
+    setRowValue(ActionField.CURRENT_VALUE_FIELD.getField(), currentCapacity);
+    setRowValue(ActionField.DESIRED_VALUE_FIELD.getField(), desiredCapacity);
```
vigyasharma, line 132: These, along with the `RESOURCE_TYPE_FIELD` are very specific to resource tuning actions. The schema wont scale to other types of actions. Instead, how about we just have a `summary` column in the table, which takes a json for each action's summary. The action can define its summary in whatever terms are suitable to it. 

Additionally, each ActionBuilder can have accept a `summary(Object obj)` field that can parse the summary to recreate its corresponding action (to allow consumers of this table to recreate actions from generic summary jsons).

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/PersistableAction.java
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * <p>Table name : Action
+ *
+ * <p>schema :
+ * | ID(primary key) |     action_name     | resource_type | resource_metric | node_id |
+ * |      1          | ModifyQueueCapacity |      1        |       5         |  node1  |
+ * | node_ip   | actionable | cool_off_period | current_value | desired_value | additional_json |
+ * | 127.0.0.1 |  true      |  300            |   100         |  150          |                 |
+ */
+public abstract class PersistableAction extends SuppressibleAction implements PersistableObject {
```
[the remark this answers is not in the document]
vigyasharma: In fact, consider moving all this logic to an `ActionListener` instead

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
```diff
@@ -50,22 +53,38 @@ public Publisher(int evalIntervalSeconds, Collator collator) {
...
     for (Action action : decision.getActions()) {
       if (coolOffDetector.isCooledOff(action) && !flipFlopDetector.isFlipFlop(action)) {
         flipFlopDetector.recordAction(action);
         coolOffDetector.recordAction(action);
         for (ActionListener listener : actionListeners) {
           listener.actionPublished(action);
         }
+        if (action instanceof PersistableAction) {
+          persistableActions.add((PersistableAction) action);
+        }
```
vigyasharma, line 78: As mentioned above, let's create a `Plugin` that will implement `ActionListener` to persist actions to sql. Then we'll add it to `actionListeners` above and this will get handled implicitly.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/Publisher.java
```diff
@@ -50,22 +53,38 @@ public Publisher(int evalIntervalSeconds, Collator collator) {
...
       }
     }
-    return new EmptyFlowUnit(Instant.now().toEpochMilli());
+    long timestamp = clock.millis();
+    if (!persistableActions.isEmpty()) {
+      return new PublisherEventFlowUnit(timestamp, new PublisherEvent(name(), timestamp, persistableActions));
```
vigyasharma, line 83: Why is this needed?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/flow_units/PublisherEventFlowUnit.java
```diff
@@ -0,0 +1,52 @@
...
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.persistence.PersistorBase;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.persistence.TableInfo;
+import java.sql.SQLException;
+
+/**
+ * A flowunit that carries PublisherEvent to downstream. The purpose of creating this flowunit is
+ * to leverage the persistor in RCA framework to persist PublisherEvent and its nested actions.
+ * This is because only flowunit can be persisted into SQL in RCA framework.
+ */
+public class PublisherEventFlowUnit extends GenericFlowUnit implements PersistableObject {
```
vigyasharma, line 18: Instead of persisting via the flow units mechanism, let's just use the `ActionListener` to persist them. Then we don't need a separate flowUnit for publisher.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/PublisherEvent.java
```diff
@@ -0,0 +1,128 @@
...
+ * By joining one PublisherEvent with the Action table, we can map one decision(here it is PublisherEvent)
+ * to a list of actions that were received and published as a batch from Collator
+ *
+ * <p>Table name : PublisherEvent
+ *
+ * <p>schema :
+ * | ID(primary key) | timestamp | event_name | number_of_actions |
+ * |      1          | 123       | Publisher  |       5           |
+ */
+public class PublisherEvent implements PersistableObject {
```
vigyasharma, line 46: This too can move to the action listener plugin which persists values to db

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/PublisherEvent.java
```diff
@@ -0,0 +1,128 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jooq.Field;
+import org.jooq.impl.DSL;
+
+/**
+ * A persistable object that persists all actionable actions published by Publisher
+ * This object will be inserted as a tuple in "PublisherEvent" table. And any actionable actions will
+ * be persisted separately in Action table and linked to this tuple via foreign key.
+ * By joining one PublisherEvent with the Action table, we can map one decision(here it is PublisherEvent)
```
vigyasharma, line 37: I think we only need to persist actions published by publisher. Do you see a need to store all actions that we create irrespective of whether or not they are published?

PR #388, opened 2020-08-19
Move RcaConf into AppContext

rguo-aws:
*Fixes #:*

*Description of changes:*
Move RcaConf into AppContext. The current rcaconf object can only be read from vertices in RCA graph. By moving this into AppContext, we can read it from any other places(action object etc.) 

*Tests:*


vigyasharma:
Are there any other changes depending on this PR?

PR #389, opened 2020-08-20
Use absolute path for configFilePath

sruti1312:
*Description of changes:*
Use absolute path for configFilePath instead of relativePath. 
In case file does not exist, pa does not startup.

*Tests:*
Spun a docker and tested.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
@@ -107,21 +107,26 @@ public class PerformanceAnalyzerApp {
         (MetricsConfiguration.CONFIG_MAP.get(StatsCollector.class).samplingInterval) / 2,
         TimeUnit.MILLISECONDS);
     PluginSettings settings = PluginSettings.instance();
-    StatsCollector.STATS_TYPE = "agent-stats-metadata";
-    METRIC_COLLECTOR_EXECUTOR.addScheduledMetricCollector(StatsCollector.instance());
-    StatsCollector.instance().addDefaultExceptionCode(StatExceptionCode.READER_RESTART_PROCESSING);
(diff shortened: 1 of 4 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
```diff
@@ -107,21 +107,26 @@ public static void main(String[] args) {
...
+          new GRPCConnectionManager(settings.getHttpsEnabled());
+      final ClientServers clientServers = createClientServers(connectionManager, appContext);
+      startErrorHandlingThread(THREAD_PROVIDER, exceptionQueue);
+
+      startReaderThread(appContext, THREAD_PROVIDER);
+      startGrpcServerThread(clientServers.getNetServer(), THREAD_PROVIDER);
+      startWebServerThread(clientServers.getHttpServer(), THREAD_PROVIDER);
+      startRcaTopLevelThread(clientServers, connectionManager, appContext, THREAD_PROVIDER);
+    } else {
+      LOG.error("Performance analyzer app stopped due to invalid config status.");
```
vigyasharma, line 128: Does the process also exit with error at this point or is it just stalled? We should exit with error.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/PerformanceAnalyzerApp.java
```diff
@@ -107,21 +107,26 @@ public static void main(String[] args) {
         (MetricsConfiguration.CONFIG_MAP.get(StatsCollector.class).samplingInterval) / 2,
         TimeUnit.MILLISECONDS);
```
sidheart, line 108: Move lines 105-108 into it if (ConfigStatus...) block so that we don't start any threads if we don't have a valid config. Can you also return -1 or some error code in that case and verify that the program exits when the config is invalid?

You should be able to do this with a simple test that executes PerformanceAnalyzerApp.main(...) and checks that the return code is the error code.

PR #400, opened 2020-08-27
Improve WaitFor utility logic

sidheart:
*Fixes #:* N/A

*Description of changes:*
- WaitFor uses Instant.now() instead of System.currentTimeMillis() to
get a system independent time measurement
- WaitFor now has a fixed wait before retesting the provided callable.
It also checks the callable one final time before throwing an exception
- A 1s WaitFor() call in WireHopperTest has been increased to 10s

*Tests:* N/A

### DIFF
--- a/src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/util/WaitFor.java
+++ b/src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/util/WaitFor.java
@@ -32,16 +32,17 @@ public class WaitFor {
     public static void waitFor(Callable<Boolean> task, long maxWait, TimeUnit unit, String message)
         throws Exception {
         long maxWaitMillis = TimeUnit.MILLISECONDS.convert(maxWait, unit);
-        Instant start = Instant.now();
-        while (!task.call() && maxWaitMillis >= 0) {
-            Instant finish = Instant.now();
-            maxWaitMillis -= Duration.between(start, finish).toMillis();
-            start = finish;
+        do {
+            Instant start = Instant.now();
+            if (task.call()) {
+                break;
+            }
+            maxWaitMillis -= Duration.between(start, Instant.now()).toMillis();
             // Wait for a fixed amount before calling the task again
-            Thread.sleep(500L);
-        }
+            Thread.sleep(100L);
+        } while (maxWaitMillis >= 0);
         // Check the task one last time before throwing an exception
-        if (!task.call() && maxWaitMillis < 0) {
+        if (!task.call()) {
             throw new TimeoutException(message);
         }
     }
--- a/src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/util/WaitFor.java
+++ b/src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/util/WaitFor.java
@@ -35,11 +35,11 @@ public class WaitFor {
         do {
             Instant start = Instant.now();
             if (task.call()) {
-                break;
+                return;
             }
-            maxWaitMillis -= Duration.between(start, Instant.now()).toMillis();
             // Wait for a fixed amount before calling the task again
             Thread.sleep(100L);
+            maxWaitMillis -= Duration.between(start, Instant.now()).toMillis();
         } while (maxWaitMillis >= 0);
         // Check the task one last time before throwing an exception
         if (!task.call()) {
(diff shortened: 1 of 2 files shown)


src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/util/WaitFor.java
```diff
@@ -16,16 +18,31 @@
...
+
+    /**
+     * Waits at most the specified time for the given task to evaluate to true
+     * @param task The task which we hope evaluates to true before the time limit
+     * @param maxWait The max amount of time to wait for the task to evaluate for true
+     * @param unit The time unit of the maxWait parameter
+     * @param message The exception message to print if waitFor times out
+     * @throws Exception If the time limit expires before the task evaluates to true
+     */
+    public static void waitFor(Callable<Boolean> task, long maxWait, TimeUnit unit, String message)
```
aditjind, line 32: Can we do it in the same function? Do we need method overloading for a single hardcoded error string?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/util/WaitFor.java
```diff
@@ -16,16 +18,31 @@
...
+            Instant finish = Instant.now();
+            maxWaitMillis -= Duration.between(start, finish).toMillis();
+            start = finish;
+            // Wait for a fixed amount before calling the task again
+            Thread.sleep(500L);
         }
-        if (maxWaitMillis < 0) {
-            throw new TimeoutException("WaitFor timed out before task evaluated to true");
+        // Check the task one last time before throwing an exception
+        if (!task.call() && maxWaitMillis < 0) {
```
yojs, line 44: Maybe a `do-while` is a cleaner way to ensure we don't miss out on the last `500` millis ?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/util/WaitFor.java
```diff
@@ -16,16 +18,31 @@
...
+        Instant start = Instant.now();
         while (!task.call() && maxWaitMillis >= 0) {
-            curTime = System.currentTimeMillis();
-            maxWaitMillis -= (curTime - pollTime);
-            pollTime = curTime;
+            Instant finish = Instant.now();
+            maxWaitMillis -= Duration.between(start, finish).toMillis();
+            start = finish;
+            // Wait for a fixed amount before calling the task again
+            Thread.sleep(500L);
```
yojs, line 41: Because this is just a test, its okay to burn CPU. If we want to be conscious, maybe 1 ms might be good enough ?

sidheart: When I was debugging I found for a particular test method that this routine failed even though at the following instruction the condition was met. This may have had something to do with the calling thread hogging resources.

I've reduced the sleep to 100ms which is extremely conservative but will still allow us to leverage the utility of WaitFor.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/util/WaitFor.java
sidheart: Done

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/util/WaitFor.java
sidheart: You're right, right now this isn't being used anywhere. However, I it's valuable for developers to be able to print their own messages context if their routine fails.

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/util/WaitFor.java
```diff
@@ -16,16 +18,32 @@
...
+            // Wait for a fixed amount before calling the task again
+            Thread.sleep(100L);
+        } while (maxWaitMillis >= 0);
+        // Check the task one last time before throwing an exception
+        if (!task.call()) {
+            throw new TimeoutException(message);
         }
     }
-}
+}
```
sruti1312, line 49: nit: new line

sruti1312 (dismissed):

yojs (approved):

sruti1312 (approved):

PR #402, opened 2020-08-31
Add support for action configs

vigyasharma:
**Fixes #:** #403 #404 #405 

**Description:** This change allows action level configurations to be provided via config file - `rca.conf`.

Additional code refactoring includes:
 - `Config` objects to encapsulate the an entire configuration. These allow config key, value, default value and validation functions to be stored in a single object.
 - Support for validating config values via predicate functions.
 - `NestedConfig` to recursively nest through a map and support multi-level nested configurations. Current support only allowed for three nested levels.
 - Remove cache bound configurations from `DeciderConfig`
 - Allow skipping muted configurations from `rca.conf`
 - Changes in `RcaConf` to allow passing config jsons as strings for tests.
 - Ignore flaky test `TimeExpiringSetTest.testExpiringIteration`

**Tests:** Added unit tests for all components

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
@@ -20,7 +20,9 @@ import static com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.
 import static com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cache.CacheUtil.MB_TO_BYTES;
 
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.AppContext;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.configs.CacheActionConfig;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.ResourceEnum;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.core.RcaConf;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.NodeKey;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.util.NodeConfigCacheReaderUtil;
 import java.util.Collections;
@@ -40,7 +42,6 @@ public class ModifyCacheMaxSizeAction extends SuppressibleAction {
 
   private final NodeKey esNode;
   private final ResourceEnum cacheType;
-  private final AppContext appContext;
 
   private final long desiredCacheMaxSizeInBytes;
   private final long currentCacheMaxSizeInBytes;
@@ -58,8 +59,6 @@ public class ModifyCacheMaxSizeAction extends SuppressibleAction {
     super(appContext);
     this.esNode = esNode;
     this.cacheType = cacheType;
-    this.appContext = appContext;
-
     this.desiredCacheMaxSizeInBytes = desiredCacheMaxSizeInBytes;
     this.currentCacheMaxSizeInBytes = currentCacheMaxSizeInBytes;
     this.coolOffPeriodInMillis = coolOffPeriodInMillis;
@@ -70,8 +69,8 @@ public class ModifyCacheMaxSizeAction extends SuppressibleAction {
       final NodeKey esNode,
       final ResourceEnum cacheType,
       final AppContext appContext,
-      double upperBoundThreshold) {
-    return new Builder(esNode, cacheType, appContext, upperBoundThreshold);
+      final RcaConf conf) {
+    return new Builder(esNode, cacheType, appContext, conf);
   }
 
   @Override
@@ -143,7 +142,7 @@ public class ModifyCacheMaxSizeAction extends SuppressibleAction {
     private final ResourceEnum cacheType;
     private final NodeKey esNode;
     private final AppContext appContext;
-    private double upperBoundThreshold;
+    private final RcaConf rcaConf;
 
     private long stepSizeInBytes;
     private boolean isIncrease;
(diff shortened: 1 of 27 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
```diff
@@ -153,34 +158,46 @@ public ResourceEnum getCacheType() {
...
     private Long heapMaxSizeInBytes;
+    private final long upperBoundInBytes;
+    private final long lowerBoundInBytes;
 
     private Builder(
         final NodeKey esNode,
         final ResourceEnum cacheType,
         final AppContext appContext,
-        double upperBoundThreshold) {
+        final RcaConf conf) {
```
rguo-aws, line 168: maybe it would be better to pass CacheActionConfig in builder ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
```diff
@@ -193,33 +180,21 @@ public Builder stepSize(int stepSize) {
       return this;
     }
 
-    public Builder upperBound(int upperBound) {
```
rguo-aws, line 196: I think we will need those two methods. The JVM decider will try to override upperbound for caches / queues

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/NestedConfig.java
```diff
@@ -0,0 +1,49 @@
...
+        LOG.error("rca.conf contains invalid value for key: [{}], trace : [{}]", key, e.getMessage());
+      }
+    }
+  }
+
+  public String getKey() {
+    return key;
+  }
+
+  public Map<String, Object> getValue() {
```
rguo-aws, line 46: Add nullable annotation to this function ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/configs/CacheActionConfig.java
```diff
@@ -0,0 +1,128 @@
...
+
+  private static class FieldDataCacheConfig implements ThresholdConfig<Double> {
+
+    private Config<Double> fieldDataCacheUpperBound;
+    private Config<Double> fieldDataCacheLowerBound;
+
+    public FieldDataCacheConfig(NestedConfig cacheSettingsConfig) {
+      NestedConfig fieldDataCacheConfig = new NestedConfig("fielddata", cacheSettingsConfig.getValue());
+      fieldDataCacheUpperBound = new Config<>("upper-bound", fieldDataCacheConfig.getValue(),
+          0.4, (s) -> (s > 0), Double.class);
```
rguo-aws, line 89: we should define the default value as a static variable ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/configs/QueueActionConfig.java
```diff
@@ -0,0 +1,127 @@
...
+
+  private static class SearchQueueConfig implements ThresholdConfig<Integer> {
+
+    private Config<Integer> searchQueueUpperBound;
+    private Config<Integer> searchQueueLowerBound;
+
+    public SearchQueueConfig(NestedConfig queueSettingsConfig) {
+      NestedConfig searchQueueConfig = new NestedConfig("search", queueSettingsConfig.getValue());
+      searchQueueUpperBound = new Config<>("upper-bound", searchQueueConfig.getValue(),
+          3000, (s) -> (s >= 0), Integer.class);
```
rguo-aws, line 88: same as above

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyCacheMaxSizeAction.java
vigyasharma: That was the original thought. But I went with this approach for the following reasons -

1. Consumers should not have to handle which config to pass to which object. The object should be able to self extract its own relevant config. 
2. Wanted to keep doors open for the action to consume configs outside of `CacheActionConfig` if needed in future.

Passing a subset config only makes it harder to do things with the config going forward (including refactoring). e.g. deciders holding only `DeciderConfig` made it impossible to pass action specific config to actions. Just passing the parent everywhere and letting consumers extract what they need made sense to me.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
vigyasharma: I have a fn to override to max and min bounds, and another to increase/decrease. I think it is too much for the decider to calculate and come up with exact values in between that it must override to. Let me know why you think otherwise.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/core/NestedConfig.java
vigyasharma: Done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/ModifyQueueCapacityAction.java
rguo-aws: the benefit of exposing upper/lower bound to decider is it allows decider to tune some certain action within a sub range of the entire {lower, upper} range. Let me use JVM decider as an example.
So the fielddata cache has a range of 0.02 - 0.4 but in JVM decider, if we only want to tune the fielddata cache within 0.2 - 0.4, the decider can override the builder to do so.

vigyasharma: We decided to expose this setter, I'll make the required changes. Documenting our reasoning here:

There can be several reasons for high old heap usage, cache and queues cover only a subset of them. Currently however, these are the only ones we are looking at. The heap health decider bucketizes an unhealthy heap into 3 levels: `level-1 unhealthy`, `level-2 unhealthy`, and `level-3 unhealthy`; with level-3 being where heap is worst hit.

At **level-1 unhealthy** (heap usage ~60-70%), we want to downsize queues and caches, but to a soft lower bound, not to the absolute hard lower bound. This is so that we don't over-penalize the only resources we are able to tune.

At **level-3 unhealthy** (heap usage >~90%), the same argument can hold, but now we know for sure that there are several other objects contesting for heap space, which is causing the 90% heap usage. So we tune down caches and queues to as much as possible, which is the hard lower bound provided by `setDesiredCacheMaxSizeToMin()`.

As a side note, we should separately see if these soft level based lower bounds should be present within deciders or within the `*ActionConfig` objects.

vigyasharma: We should pick this up as a separate PR. The soft and hard bounds should be attributes of action config objects itself, so that they can have corresponding overrides in rca.conf, and we don't maintain and pass separate bound values from deciders. 

Instead of exposing the `lowerBound()` setter, we should expose a `useSoftLowerBound()` setter, with soft lower bound being read from rca.conf.

PR #408, opened 2020-09-01
Change SubscribeResponseHandler#onError to log at WARN level

sidheart:
*Fixes #:* N/A

*Description of changes:*
- The Performance Analyzer IT Framework throws an exception if it sees
error logs. A unit test expressly calls
SubscribeResponseHandler#onError, which writes an error log. Therefore,
if this unit test runs before the RCA IT, we'll get an Exception for no
good reason. This commit changes the log level to WARN, which avoids
this behavior.

*Tests:* SubscribeResponseHandlerTest


sidheart:
We can solve this in a more reasonable way

PR #413, opened 2020-09-04
Persist actions in sqlite

sruti1312:
*Fixes #:*

*Description of changes:*
Persist actions in sqlite

*Tests:*
unit test and IT


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
```diff
@@ -64,7 +64,7 @@
 import org.jooq.exception.DataAccessException;
 import org.jooq.impl.DSL;
 
-class SQLitePersistor extends PersistorBase {
+public class SQLitePersistor extends PersistorBase {
```
yojs, line 67: We might not want to make it public. The idea is all users should use the public interface rather than actual DB implementor. Would you mind your usecase for this change ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/actions/ActionsSummary.java
```diff
@@ -0,0 +1,96 @@
...
+            String actionName, long resourceValue, String id, String ip, boolean actionable, long coolOffPeriod, long timestamp) {
+        this.actionName = actionName;
+        this.resourceValue = resourceValue;
+        this.id = id;
+        this.ip = ip;
+        this.actionable = actionable;
+        this.coolOffPeriod = coolOffPeriod;
+        this.timestamp = timestamp;
+    }
+
```
yojs, line 40: Do we really need this as we have setters/getters for individual fields ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/response/RcaResponse.java
```diff
@@ -137,6 +137,7 @@ public JsonElement toJson() {
   public GenericSummary buildNestedSummary(String summaryTable, Record record) {
     GenericSummary ret = null;
     if (summaryTable.equals(HotClusterSummary.HOT_CLUSTER_SUMMARY_TABLE)) {
+
```
yojs, line 140: remove ?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/plugins/PublisherEventsPersistorTest.java
```diff
@@ -0,0 +1,110 @@
...
+        public boolean isMuted() {
+            return false;
+        }
+
+        @Override
+        public ResourceEnum getResource() {
+            return ResourceEnum.FIELD_DATA_CACHE;
+        }
+    }
+}
```
yojs, line 110: new line in the end ?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/integTests/framework/Cluster.java
```diff
@@ -273,6 +273,10 @@ public JsonElement getAllRcaDataOnHost(HostTag hostTag, String rcaName) {
     return tagToHostMapping.get(hostTag).getDataForRca(rcaName);
   }
 
+  public <T> T getAllClassDataOnHost(HostTag hostTag, Class<T> className) throws Exception {
```
yojs, line 276: Maybe we can name it something like `read/constructObjectFromDB` ?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/integTests/framework/Host.java
```diff
@@ -368,6 +369,10 @@ public JsonElement getDataForRca(String rcaName) {
     return obj;
   }
 
+  public <T> T getDataForClass(Class<T> className) throws Exception {
+    return this.rcaController.getPersistenceProvider().read(className);
```
yojs, line 373: ditto as the comment on cluster.java

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/plugins/PublisherEventsPersistor.java
```diff
@@ -0,0 +1,71 @@
...
+    public static final String NAME = "publisher_events_persistor_plugin";
+    private static final Logger LOG = LogManager.getLogger(PublisherEventsPersistor.class);
+
+    private static Persistable persistable = null;
+
+    @Override
+    public void actionPublished(final Action action) {
+        long timestamp = Instant.now().toEpochMilli();
+        if (persistable != null) {
+            LOG.info("Action: [{}] published to persistor publisher.", action.name());
```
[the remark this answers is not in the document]
sruti1312: Modified to debug and added a metric to track it

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/plugins/PublisherEventsPersistor.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.logging.log4j.Logger;
+
+/**
+ * A listener that persists all actions published by the publisher to rca.sqlite
+ */
+public class PublisherEventsPersistor extends Plugin implements ActionListener {
+    public static final String NAME = "publisher_events_persistor_plugin";
+    private static final Logger LOG = LogManager.getLogger(PublisherEventsPersistor.class);
+
+    private static Persistable persistable = null;
```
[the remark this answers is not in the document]
sruti1312: Right. Modified

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/actions/ActionsSummary.java
sruti1312: Added it as it makes the code cleaner. Modified to use getters to set the values

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/integTests/framework/Cluster.java
sruti1312: Using constructObjectFromDB

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/integTests/framework/Host.java
sruti1312: Modified using constructObjectFromDB

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/plugins/PublisherEventsPersistorTest.java
sruti1312: Done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/response/RcaResponse.java
sruti1312: Done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
sruti1312: Using @VisibleForTesing. I am creating a sqlitePersistor for my test purposes.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/plugins/PublisherEventsPersistor.java
```diff
@@ -0,0 +1,71 @@
...
+                            action.getResource().getNumber(),
+                            nodeKey.getNodeId().toString(),
+                            nodeKey.getHostAddress().toString(),
+                            action.isActionable(),
+                            action.coolOffPeriodInMillis(),
+                            timestamp);
+                    try {
+                        persistable.write(actionsSummary);
+                    } catch (Exception e) {
+                        LOG.error("Unable to write publisher events to sqlite", e);
```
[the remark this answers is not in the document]
sruti1312: Added a metric

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/SQLitePersistor.java
sruti1312: I was using the sqlitePersistor for my test purposes. Moved the test, I no more require it there. Removed this

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/actions/ActionsSummary.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * <p>Table name : ActionsSummary
+ *
+ * <p>schema :
+ * | id(primary key) |     actionName       |  resourceValue    |  timestamp        |   nodeId      |
+ * |      1          | ModifyQueueCapacity  |      1            |  1599257910923    |   node1       |
+ * | nodeIp     | actionable | coolOffPeriod
+ * | 127.0.0.1  |  true      |  300
+ */
+public class ActionsSummary {
```
vigyasharma, line 19: Actions can be wide ranging and diverse, from resource tuning ones like queues and caches, to cluster management actions like shard movements. We should strive to keep that freedom in our action objects. This is why the `Action` interface was kept generic enough to accommodate for these different kind of actions - only items that each action must have should be imposed there. We should do the same with summaries.

Having an `ActionSummary` class will impose a rigid structure that actions must adhere to. Instead, we can have a simple factory kind of model for generating actions from summaries here:

Each action exposes two methods: 
```java
/* returns action summary as a json string. 
* We can also use jsonElement but string is just 
*/
public String summary(); more generic. The string would still be json. 

// this one is in each action, not in the interface
// Takes action summary as json string and constructs action object from it
public static Action fromSummary(String jsonRepr); 
```

Now each action has its own implementations for these methods, and thus its own impl for serializing and deserializing its contents. The validators can construct an `Action` object from json responses by calling `fromSummary` and validate if these objects are the expected ones.
e.g.
```java
// throws exception if summary is not parse-able to queue action
ModifyQueueCapacityAction action = ModifyQueueCapacityAction.fromSummary(response); 
assertEquals(ModifyQueueCapacityAction.NAME, action.name());
```

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/Action.java
```diff
@@ -48,4 +49,6 @@
 
   /** Returns if this action is explicitly muted through configuration */
   boolean isMuted();
+
+  ResourceEnum getResource();
```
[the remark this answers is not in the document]
vigyasharma: I dont think we need this in the interface. This method can just exist in the queue and cache actions (i think we already have `getThreadPool()` there. The validators would know exactly which action they are expecting and can call these methods directly (see subsequent comment on action summaries for more context).

vigyasharma (changes requested):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/metrics/ExceptionsAndErrors.java
```diff
@@ -37,7 +37,12 @@
   /**
    * When persisting an RCA throws an exception.
    */
-  EXCEPTION_IN_PERSIST("ExceptionInPersist", "namedCount", Statistics.NAMED_COUNTERS);
+  EXCEPTION_IN_PERSIST("ExceptionInPersist", "namedCount", Statistics.NAMED_COUNTERS),
+
+  /**
+   * When persisting an action, the persistable throws an exception when it is unable to write to DB.
+   */
+  EXCEPTION_IN_WRITE_TO_SQLITE("ExceptionInWriteToSqlite", "namedCount", Statistics.NAMED_COUNTERS);
```
yojs, line 45: Can we call out the difference between `EXCEPTION_IN_WRITE_TO_SQLITE` and `EXCEPTION_IN_PERSIST`and why we require both ? 
The thing with named counters is you can bucketize over the same enum. So, say `EXCEPTION_IN_PERSIST` metric is reported when `EX1` is thrown and also when `EX2` is thrown. You can pass that along and in the reported metrics you will see both buckets with counts.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/metrics/RcaRuntimeMetrics.java
```diff
@@ -43,7 +43,12 @@
    * Metric tracking if RCA is enabled or disabled. We write a 0 if RCA is disabled and 1 if it is
    * enabled.
    */
-  RCA_ENABLED("RcaEnabled", "count", Collections.singletonList(Statistics.SAMPLE));
+  RCA_ENABLED("RcaEnabled", "count", Collections.singletonList(Statistics.SAMPLE)),
+
+  /**
+   * Metric tracking the actions published by the publisher that are persisted in sqlite.
+   */
+  ACTIONS_PUBLISHED("ActionsPublished", "count", Collections.singletonList(Statistics.NAMED_COUNTERS));
```
yojs, line 51: Maybe we want to s/count/namedCount/ . Because this API supports it, we should get a counter for each proposed action ?

This is great that we are adding these metrics. It will give as a visibility into the tuning framework. Thanks for adding !

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PublisherEventsPersistor.java
```diff
@@ -0,0 +1,73 @@
...
+    public PublisherEventsPersistor(final Persistable persistable) {
+        this.persistable = persistable;
+    }
+
+    public void persistAction(final Action action) {
+        long timestamp = Instant.now().toEpochMilli();
+        if (persistable != null) {
+            LOG.debug("Action: [{}] published to persistor publisher.", action.name());
+            PerformanceAnalyzerApp.RCA_RUNTIME_METRICS_AGGREGATOR.updateStat(
+                    RcaRuntimeMetrics.ACTIONS_PUBLISHED, this.name(), 1);
```
yojs, line 45: Should we push these lines inside the if block ? If there are no actions, we may not want the counter to the incremented, right ? Do you just wan to track how many times this method is called ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/actions/ActionsSummary.java
yojs: +1

Maybe we should have separate Action classes for each action type to let them have the freedom to decide on the schema ?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/integTests/framework/runners/RcaItRunnerBase.java
```diff
@@ -267,6 +267,12 @@ private void applyMethodLevelAnnotationOverrides(Method method) throws Exception
                 failedValidations.add(validator.getClass());
               }
               break;
+            case SQLITE:
+              successful = validator.check(testApi.constructObjectFromDBOnHost(expect.on(), rca));
+              if (!successful) {
+                failedValidations.add(validator.getClass());
+              }
```
yojs, line 274: Because that piece is common, do we wan to move it to the else block of `if (successful) {` ?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/integTests/framework/annotations/AExpect.java
```diff
@@ -53,7 +53,8 @@
    * Currently supported places to look for RCA outputs are the SQLite file or by hitting the rest endpoint.
    */
   enum Type {
-    REST_API
+    REST_API,
+    SQLITE
```
yojs, line 57: Thank you for extending the API ! :)

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/PublisherEventsPersistor.java
sruti1312: If block is for impactedNodes. This method will be called only when an action is present. We can retail this outside.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/metrics/ExceptionsAndErrors.java
sruti1312: I think we can use persist for both. As it is named counter, we will have a counter for each one.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/metrics/RcaRuntimeMetrics.java
sruti1312: Right. Changed this.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/persistence/actions/ActionsSummary.java
sruti1312: I am adding summary to the ActionSummary class (wrapper). Presently all the actions contains this summary. Will create a separate PR for adding fromSummary. 

About creating separate action and allowing them to decide the schema, might be some more thought. I think we start with this and add more fields as use cases increase.

PR #414, opened 2020-09-08
Add unit tests for JVM decider

rguo-aws:
*Fixes #:*
#409 

*Description of changes:*
1. Add unit test for JVM decider and relevant decision making modules
2. Fix minor bugs found when writing unite tests

*Tests:*


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/configs/DeciderConfig.java
```diff
@@ -26,7 +24,8 @@
  *     // Decreasing order of priority for the type of workload we can expect on the cluster.
  *     // Priority order in the list goes from most expected to the lease expected workload type.
  *     "workload-type": {
- *       "priority-order": ["ingest", "search"]
+ *       "prefer-ingest": true,
+ *       "prefer-search": false
```
sruti1312, line 28: What happens when there are more than 2 workload type. If we dont want them in a list, can we use numbers to denote their priority? Also can you update the comment if we are changing it

rguo-aws: I actually had a discussion with Vigya on this in my previous PR. The problem with setting workload as a list is we have to maintain the string parsing logic inside the Config object and the priority list does not allow customer to specify the scenario if both search/ingest are equally important to customer or their simply don't have any preference. 
So we created two options here instead so that customer can explicitly specific whether search/ingest is preferable. 
The comments has been updated in the WorkLoaderConfig.java

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/configs/CacheActionConfig.java
```diff
@@ -68,7 +68,7 @@ public CacheActionConfig(RcaConf conf) {
     cacheSettingsConfig = new NestedConfig("cache-settings", actionConfig);
     fieldDataCacheConfig = new FieldDataCacheConfig(cacheSettingsConfig);
     shardRequestCacheConfig = new ShardRequestCacheConfig(cacheSettingsConfig);
-    totalStepCount = new Config<>(TOTAL_STEP_COUNT_CONFIG_NAME, cacheSettingsConfig.getValue(),
+    totalStepCount = new Config<>(TOTAL_STEP_COUNT_CONFIG_NAME, actionConfig,
```
sruti1312, line 71: step count will be a part of cacheSettingConfig right? Is there a reason why we are moving to actionConfig?
Will it be common parameter for cache and queue?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/configs/QueueActionConfig.java
```diff
@@ -67,7 +67,7 @@ public QueueActionConfig(RcaConf conf) {
     queueSettingsConfig = new NestedConfig("queue-settings", actionConfig);
     searchQueueConfig = new SearchQueueConfig(queueSettingsConfig);
     writeQueueConfig = new WriteQueueConfig(queueSettingsConfig);
-    totalStepCount = new Config<>(TOTAL_STEP_COUNT_CONFIG_NAME, queueSettingsConfig.getValue(),
+    totalStepCount = new Config<>(TOTAL_STEP_COUNT_CONFIG_NAME, actionConfig,
```
sruti1312, line 70: same as above

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/configs/DeciderConfig.java
```diff
@@ -26,7 +24,8 @@
  *     // Decreasing order of priority for the type of workload we can expect on the cluster.
  *     // Priority order in the list goes from most expected to the lease expected workload type.
```
sruti1312, line 25: We might need to update this comment right?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/configs/QueueActionConfig.java
```diff
@@ -36,7 +36,7 @@
  *       "total-step-count": 20,
```
sruti1312, line 36: This is confusing. I notice you have changed to use actionConfig. In that case this should be outside queue-settings block right?

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/DeciderActionParser.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.Action;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.ModifyCacheMaxSizeAction;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.ModifyQueueCapacityAction;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.ResourceEnum;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DeciderActionParser {
```
sruti1312, line 26: As we are using this in test, can we rename to DeciderActionParserUtil or put it under util folder?
This can avoid confusion between src files and test files

sruti1312 (approved):

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/DeciderActionParser.java
sidheart: +1

sidheart (approved):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/configs/DeciderConfig.java
rguo-aws: done

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/configs/QueueActionConfig.java
rguo-aws: this is the step count to divide the range of queue size into steps. We can specify different step count in cache if we want to have different granularity

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/DeciderActionParser.java
rguo-aws: good point. Let me rename this class

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/configs/CacheActionConfig.java
rguo-aws: step count is indeed part of the cacheSettingConfig. But here we need to pass the parent Map<> object as a parameter in the constructor of the Config class

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/configs/CacheActionConfigTest.java
```diff
@@ -105,4 +105,48 @@ public void testInvalidCacheType() throws Exception {
...
+      + "}";
+    RcaConf conf = new RcaConf();
+    conf.readConfigFromString(configStr);
+    CacheActionConfig cacheActionConfig = new CacheActionConfig(conf);
+    assertEquals(0.03, cacheActionConfig.getStepSize(ResourceEnum.FIELD_DATA_CACHE), 0.0001);
+    assertEquals(0.0035, cacheActionConfig.getStepSize(ResourceEnum.SHARD_REQUEST_CACHE), 0.0001);
+    configStr =
+        "{"
+          + "\"action-config-settings\": { "
+              + "\"total-step-count\": 5,"
```
sruti1312, line 134: As per your comment 
```step count is indeed part of the cacheSettingConfig. But here we need to pass the parent Map<> object as a parameter in the constructor of the Config class```
In that case this step count might need to be inside the cache-settings block right?

sidheart (approved):

sruti1312 (approved):

Issue #426, opened 2020-09-17
PowermockRunner is currently unusable


ktkrg:
Mocking static classes using Powermock is currently not possible in the code base as it causes Log4j to fail during initialization of many classes. We will need to get this working if we want to enable some reader tests in our codebase.

PR #427, opened 2020-09-17
Re enable config overrides test

ktkrg:
*Fixes #:*
#422 

*Description of changes:*
This change makes the two methods: `updateAllRcaConfs` and `updateRcaConf` non static even though they don't mutate any state so that the unit tests for ConfigOverridesApplier can be enabled again.

While it is easy to refactor it out into a static helper, mocking static objects is currently not a simple fix in our codebase as PowerMockRunner is not ignoring log4j classes and is causing initialization issues for a lot of classes during unit testing(#426)

This is a simple fix for enabling the tests until #426 is resolved.

*Tests:*
`./gradlew build`


yojs:
I like how clearly you state why this is done the way it is done. Thanks for the follow up issue creation.

yojs (approved):

vigyasharma (approved):

PR #428, opened 2020-09-17
Add cache clear action in JVM decider

rguo-aws:
*Fixes #:*

*Description of changes:*
Add cache clear action in JVM decider. This action will be issued when the heap usage(after full GC) is above 90%. The default cool off period of this action is 30 mins

*Tests:*
created unit tests for this action type

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/CacheClearAction.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/CacheClearAction.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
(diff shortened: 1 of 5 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelThreeActionBuilder.java
```diff
@@ -117,31 +113,28 @@ private void addSearchQueueAction() {
...
 
-
-  private void registerActions() {
-    addFieldDataCacheAction();
-    addShardRequestCacheAction();
-    addSearchQueueAction();
-    addWriteQueueAction();
+  private void addCacheClearAction() {
+    CacheClearAction action = CacheClearAction
+        .newBuilder(appContext).build();
```
vigyasharma, line 122: nit: same line?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/actions/CacheClearAction.java
```diff
@@ -0,0 +1,123 @@
...
+  public List<NodeKey> impactedNodes() {
+    return impactedNodes;
+  }
+
+  @Override
+  public Map<NodeKey, ImpactVector> impact() {
+    Map<NodeKey, ImpactVector> impactedMap = new HashMap<>();
+    impactedNodes.forEach(node -> {
+          ImpactVector impactVector = new ImpactVector();
+          impactVector.decreasesPressure(HEAP);
```
vigyasharma, line 72: Will this action increase `CPU` pressure? Although, since this is an SoS action, we want it to go through and not get punted due to an existing high CPU metric. So it's probably okay to not include that as an impact (also may not always increase CPU if the cache items were not really being used).

vigyasharma (approved):

sidheart (approved):

PR #445, opened 2020-10-02
Fixing broken PA for Node to Node Encyrption clusters

khushbr:
*Fixes #:* https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/issues/444

*Description of changes:* Fixing broken PA for Node to Node Encyrption clusters

*Tests:* Existing Integ Test, verified on devstack cluster

### DIFF
--- a/pa_config/performance-analyzer.properties
+++ b/pa_config/performance-analyzer.properties
@@ -25,7 +25,7 @@ https-enabled = false
 # Setup the correct path for server certificates
 certificate-file-path = specify_path
 private-key-file-path = specify_path
-trusted-cas-file-path = specify_path
+#trusted-cas-file-path = specify_path
 
 # Setup the correct path for client certificates (by default, the client will just use the server certificates)
 #client-certificate-file-path = specify_path


sidheart (approved):

sidheart:
If this is the file we're reading in AES then this should work. Are you sure that leaving the other two lines uncommented is fine?

ktkrg (approved):

pa_config/performance-analyzer.properties
```diff
@@ -25,7 +25,7 @@ https-enabled = false
 # Setup the correct path for server certificates
 certificate-file-path = specify_path
 private-key-file-path = specify_path
-trusted-cas-file-path = specify_path
+#trusted-cas-file-path = specify_path
```
aditjind, line 28: While debugging we had to change the config file at 2 places. While it might be possible that same config file is kept at 2 places, but something we can check before merging.

vigyasharma (approved):

pa_config/performance-analyzer.properties
khushbr: Right, so we copy this file to the 2 locations.

khushbr: Right, so the same file gets copied at the 2 locations.

PR #446, opened 2020-10-02
Modify NodeConfigCollector to include more heap stats

sidheart:
Modify NodeConfigCollector to include more heap stats

*Fixes #:* N/A

*Description of changes:*
- NodeConfigCollector now reports the maximum amount of memory avaiable
in the old generation and the young generation in bytes

*Tests:* 
- NodeConfigCollectorTest

Negligible effect on overall test time

### DIFF
--- a/src/main/proto/inter_node_rpc_service.proto
+++ b/src/main/proto/inter_node_rpc_service.proto
@@ -83,7 +83,7 @@ enum MetricEnum {
   // JVM
   HEAP_USAGE = 0 [(additional_fields).name = "heap usage", (additional_fields).description = "memory usage in percentage"];
   PROMOTION_RATE = 1 [(additional_fields).name = "promotion rate", (additional_fields).description = "mb/s"];
-  MINOR_GC = 2 [(additional_fields).name = "minor gc", (additional_fields).description = "time in percentage"];
+  MINOR_GC = 2 [(additional_fields).name = "minor gc", (additional_fields).description = "minor gc pause time in ms"];
 
   // hardware
   CPU_USAGE = 3 [(additional_fields).name = "cpu usage", (additional_fields).description = "num of cores"];
@@ -101,6 +101,9 @@ enum MetricEnum {
 
   // Heap
   HEAP_MAX = 16 [(additional_fields).name = "heap max", (additional_fields).description = "max heap size in bytes"];
+
+  // GC
+  FULL_GC = 32 [(additional_fields).name = "full gc", (additional_fields).description = "full gc pause time in ms"];
 }
 
 /*
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/ResourceUtil.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/ResourceUtil.java
@@ -31,7 +31,7 @@ import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.ResourceEnum
  * such as TOTAL_THROUGHPUT and SYS_CALL_RATE
  */
 public class ResourceUtil {
-  // JVM resource
+  // JVM resources
   public static final Resource HEAP_MAX_SIZE = Resource.newBuilder()
           .setResourceEnum(ResourceEnum.HEAP)
           .setMetricEnum(MetricEnum.HEAP_MAX).build();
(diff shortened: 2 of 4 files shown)


rguo-aws (approved):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/util/NodeConfigCacheReaderUtil.java
```diff
@@ -64,4 +64,24 @@ public static Long readHeapMaxSizeInBytes(
...
+      LOG.error("Exception while reading heap max size from Node Config Cache", e);
+    }
+    return null;
+  }
+
+  public static Double readYoungGenMaxSizeInBytes(
+      final NodeConfigCache nodeConfigCache, final NodeKey esNode) {
+    try {
+      return nodeConfigCache.get(esNode, ResourceUtil.YOUNG_GEN_MAX_SIZE);
+    } catch (final IllegalArgumentException e) {
```
vigyasharma, line 82: The whole idea of throwing an exception from `nodeConfigCache.get` was to make callers aware of a missing value and **letting them decide** how to handle it. By masking it and returning `null` in a reader util, we are exposing ourselves to future bugs.
I don't understand how this is useful or better, the caller still needs to check for null values - only now the compiler will not ensure that these values are checked. So we run a greater risk of missing these values. That is the whole point of checked exceptions.
This reader util class is good in that you don't have to remember the actual `Resource` values to pass, but it should follow the same API signature. We should not silently eat these exceptions here and make callers handle a different unchecked code path.

PS: I know this is not a new artifact in this PR, it is simply extending existing code faults. I think we should fix it, if the change is big, we can create separate PRs. Either way, at least the new methods can follow the right signature.

vigyasharma: Also `nodeConfigCache.get` should throw a checked exception, not a RuntimeException (which are for critical/unrecoverable failures)

src/main/proto/inter_node_rpc_service.proto
```diff
@@ -101,6 +101,9 @@ enum MetricEnum {
 
   // Heap
   HEAP_MAX = 16 [(additional_fields).name = "heap max", (additional_fields).description = "max heap size in bytes"];
+
+  // GC
+  FULL_GC = 32 [(additional_fields).name = "full gc", (additional_fields).description = "pause time in ms"];
```
vigyasharma, line 106: It seems `MINOR_GC` is described as `"time in percentage"`. It that valid?

vigyasharma (approved):
Left some questions and best practices comments. Approving from a functional angle, pls address/respond to the comments before merging.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/util/NodeConfigCacheReaderUtil.java
sidheart: I disagree with your comment on 2 levels

1. The purpose of `NodeConfigCacheReaderUtil` is to eat the exceptions and let consumers gracefully handle nulls as you can see is being done today. This is a valid decision and seems unlikely to lead to future bugs because all consumers today are using this the same way and future consumers are likely to copy the same mechanism.

1. We don't allow storing null values in the NodeConfigCache even if the signature technically allows it. Look at the signature of put(), we can only store actual double values in the cache. We have to use a boxed type for the cache because Java. Since it isn't possible to have null values in the cache, a ReaderUtil returning null will not miss any values and will signal what we want it to signal.

Let me know if you see an issue with my reasoning.

src/main/proto/inter_node_rpc_service.proto
sidheart: Changed the description

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/rca/util/NodeConfigCacheReaderUtil.java
sidheart: I removed the functions I added to this class, my consumers down the line will use the NodeConfigCache directly

vigyasharma (approved):

rguo-aws (approved):

PR #447, opened 2020-10-02
Add the PersistableSlidingWindow class

sidheart:
Add the PersistableSlidingWindow class

*Fixes #:* N/A

*Description of changes:*
- A PersistableSlidingWindow is a sliding window that aggregates data
into buckets and can be written to and read from the disk. This makes it
suitable for retaining data over large spans of time requiring too much
memory or being susceptible to node loss.

*Tests:*
- PersistableSlidingWindowTest

- Test time is negligible

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
@@ -50,8 +50,7 @@ public class PersistableSlidingWindow extends SlidingWindow<SlidingWindowData> {
         load(pathToFile);
       }
     } catch (IOException e) {
-      LOG.error("Couldn't create file {} to perform young generation tuning", pathToFile, e);
-      throw new IllegalArgumentException("Couldn't create or read a file at " + filePath);
+      LOG.error("Couldn't load SlidingWindow data from {}", pathToFile, e);
     }
   }
 
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.aggregators;
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.persistence.FileRotate;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Iterator;
+import java.util.concurrent.TimeUnit;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.LineIterator;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * PersistableSlidingWindow is a SlidingWindow which can have its data written to and read from disk
+ */
+public class PersistableSlidingWindow extends SlidingWindow<SlidingWindowData> {
+  private static final Logger LOG = LogManager.getLogger(PersistableSlidingWindow.class);
+  private static final ObjectMapper objectMapper = new ObjectMapper();
+
+  private Path pathToFile;
+
+  public PersistableSlidingWindow(int slidingWindowSize,
+                                  TimeUnit timeUnit,
+                                  Path filePath) {
+    super(slidingWindowSize, timeUnit);
+    this.pathToFile = filePath;
+    try {
+      if (Files.exists(pathToFile)) {
+        load(pathToFile);
+      }
+    } catch (IOException e) {
+      LOG.error("Couldn't create file {} to perform young generation tuning", pathToFile, e);
+      throw new IllegalArgumentException("Couldn't create or read a file at " + filePath);
+    }
+  }
+
+  /**
+   * Loads the SlidingWindowData contained in the given path into this PersistableSlidingWindow
+   * @param path The path to the file containing the SlidingWindow data
+   * @throws IOException If there is an error reading the file
+   */
+  protected synchronized void load(Path path) throws IOException {
+    LineIterator it = FileUtils.lineIterator(path.toFile(), "UTF-8");
+    try {
+      while (it.hasNext()) {
+        String line = it.nextLine();
+        SlidingWindowData data = objectMapper.readValue(line, SlidingWindowData.class);
+        next(data);
+      }
+    } finally {
+      LineIterator.closeQuietly(it);
+    }
+  }
+
+  /**
+   * Writes the contents of this SlidingWindow into the path provided during construction
+   *
+   * <p>This write occurs in 2 stages, it writes to a temporary file, then replaces the actual data
+   * file with the contents of the temporary file
+   *
+   * @throws IOException If there is a CRUD error with the files involved
+   */
+  protected synchronized void write() throws IOException {
+    String tmpFile = pathToFile.toString() + RandomStringUtils.randomAlphanumeric(32);
+    Path tmpPath = Paths.get(tmpFile);
+    Files.createFile(Paths.get(tmpFile));
+    BufferedWriter writer = new BufferedWriter(new FileWriter(tmpFile, false));
+    Iterator<SlidingWindowData> it = windowDeque.descendingIterator();
+    while (it.hasNext()) {
+      writer.write(objectMapper.writeValueAsString(it.next()));
+      writer.write(System.lineSeparator());
+    }
+    // write to temporary file
+    writer.flush();
+    // atomic rotate
+    FileRotate.rotateFile(tmpPath, pathToFile);
+  }
+}
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/SlidingWindow.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/SlidingWindow.java
@@ -28,11 +28,13 @@ public class SlidingWindow<E extends SlidingWindowData> {
   protected final Deque<E> windowDeque;
   protected final long SLIDING_WINDOW_SIZE;
   protected double sum;
+  protected TimeUnit timeUnit;
 
   public SlidingWindow(int SLIDING_WINDOW_SIZE_IN_TIMESTAMP, TimeUnit timeUnit) {
     this.windowDeque = new LinkedList<>();
     this.SLIDING_WINDOW_SIZE = timeUnit.toSeconds(SLIDING_WINDOW_SIZE_IN_TIMESTAMP);
     this.sum = 0.0;
+    this.timeUnit = timeUnit;
   }
 
   /**
@@ -92,5 +94,9 @@ public class SlidingWindow<E extends SlidingWindowData> {
    */
   public double readSum() {
     return this.sum;
+  }
+
+  public int size() {
+    return windowDeque.size();
   }
 }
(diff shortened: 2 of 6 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  @Override
+  public void next(SlidingWindowData slidingWindowData) {
+    if (aggregatedData == null) {
+      aggregatedData = new SlidingWindowData(slidingWindowData.getTimeStamp(), slidingWindowData.getValue());
+      return;
+    }
+    long timestampDiff = slidingWindowData.getTimeStamp() - aggregatedData.getTimeStamp();
+    if (timeUnit.convert(timestampDiff, TimeUnit.SECONDS) > bucketSizeMs) {
```
vigyasharma, line 83: Is this functionally correct? IIUC, this expects `timestampDiff` to be in `SECONDS`, converts it to `timeUnit` and compares it against `bucketSizeMs` which is in milliseconds.
Is `timeUnit` guaranteed to be in ms? Is `timestampDiff` guaranteed to be in seconds?

vigyasharma (changes requested):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
```diff
@@ -0,0 +1,134 @@
...
+ * <p>The sliding window retains data for slidingWindowSizeInSeconds and aggregates values into time
+ * buckets of width bucketSizeInSeconds. For example if slidingWindowSizeInSeconds is 86400 (1 day)
+ * and bucketSizeInSeconds is 3600 (1 hr) then there will be at most 25 data points in the sliding window
+ * at any given time. It's 25 and not 24 because the oldest datapoints are removed lazily based on
+ * arrival time. This shouldn't impact most consumers of this data structure.
+ *
+ * <p>The data contained in this sliding window is effectively a time series. Consumers may extend this
+ * class and define more complex aggregations or analytics functions.
+ */
+public class PersistableSlidingWindow extends SlidingWindow<SlidingWindowData> {
```
vigyasharma, line 46: This class is doing two things, it is providing a persistable sliding window, and it is aggregating (deduping) data points within a bucketization window.
How about we split these responsibilities and have this class focus only on persisting?

Re: the bucketization piece, I have a few more comments - I suppose the motivation for this is to act like an `alarm breach monitor` which gets set when some thresholds are hit, and unset on a different criteria. This can be used in deciders then, to figure out when an action is relevant (alarm unhealthy) and irrelevant (alarm healthy).

As such, it may be read more often than an issue gets added (`next` being called) and the sliding window should get updated on each read. 

Again, we can move this to a separate PR/class and focus this one primarily on persistence as I see most of the code here pertains to persisting data.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
sidheart: Good catch, fixed and added a UT to catch this.

SlidingWindowData timestamps must be in ms (SlidingWindow assumes this). The bucket length and overall window size can be specified using any timeunit.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
sidheart: I removed the aggregation logic from this PR.

vigyasharma (dismissed):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.logging.log4j.Logger;
+
+/**
+ * PersistableSlidingWindow is a SlidingWindow which can have its data written to and read from disk
+ *
+ * <p>The sliding window retains data for slidingWindowSizeInSeconds and aggregates values into time
+ * buckets of width bucketSizeInSeconds. For example if slidingWindowSizeInSeconds is 86400 (1 day)
+ * and bucketSizeInSeconds is 3600 (1 hr) then there will be at most 25 data points in the sliding window
+ * at any given time. It's 25 and not 24 because the oldest datapoints are removed lazily based on
+ * arrival time. This shouldn't impact most consumers of this data structure.
```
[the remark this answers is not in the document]
sidheart: Updated comments in PersistableSlidingWindow

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
```diff
@@ -0,0 +1,103 @@
...
+    BufferedWriter writer = new BufferedWriter(new FileWriter(tmpFile, false));
+    Iterator<SlidingWindowData> it = windowDeque.descendingIterator();
+    while (it.hasNext()) {
+      writer.write(objectMapper.writeValueAsString(it.next()));
+      writer.write(SEPARATOR);
+    }
+    // write to temporary file
+    writer.flush();
+    // atomic rotate
+    FileRotate.rotateFile(tmpPath, pathToFile);
```
rguo-aws, line 101: do we want to make rotateFile() synchronized ?

And can we refactor FileRotate class to make it more generic instead of using static to expose its method directly ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
```diff
@@ -0,0 +1,103 @@
...
+  protected static final String SEPARATOR = "\n";
+
+  private Path pathToFile;
+
+  public PersistableSlidingWindow(int slidingWindowSize,
+                                  TimeUnit timeUnit,
+                                  Path filePath) {
+    super(slidingWindowSize, timeUnit);
+    this.pathToFile = filePath;
+    try {
```
rguo-aws, line 50: should we do this check within the constructor itself ? I think it would make more sense to perform this check during writing operation.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
```diff
@@ -0,0 +1,103 @@
...
+   * file with the contents of the temporary file
+   *
+   * @throws IOException If there is a CRUD error with the files involved
+   */
+  protected void write() throws IOException {
+    String tmpFile = pathToFile.toString() + RandomStringUtils.randomAlphanumeric(32);
+    Path tmpPath = Paths.get(tmpFile);
+    Files.createFile(Paths.get(tmpFile));
+    BufferedWriter writer = new BufferedWriter(new FileWriter(tmpFile, false));
+    Iterator<SlidingWindowData> it = windowDeque.descendingIterator();
```
ktkrg, line 93: any reason why we're writing in the reverse sequential order? Won't this bring back the sliding window in the reverse order after a load?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
sidheart: I made the write and load synchronized, we don't want to make rotateFile synchronized

FileRotate contains a lot of specific logic for logfile rotation with a date prefix. Refactoring it is out of scope. I've made minimal changes to address the thread safety issue you brought up.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
sidheart: I don't expose the load function to consumers. This class loads previous data exactly once upon construction. It doesn't need to load during write/next operations.

sidheart: The reason I create the file here is because I need to do the check before I load. The reason this logic shouldn't live inside write() is because it would waste every single instruction to check beyond the first

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.logging.log4j.Logger;
+
+/**
+ * PersistableSlidingWindow is a SlidingWindow which can have its data written to and read from disk
+ */
+public class PersistableSlidingWindow extends SlidingWindow<SlidingWindowData> {
+  private static final Logger LOG = LogManager.getLogger(PersistableSlidingWindow.class);
+  private static final ObjectMapper objectMapper = new ObjectMapper();
+
+  protected static final String SEPARATOR = "\n";
```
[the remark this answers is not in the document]
sidheart: File.separator is generally a path separator like '/'

In this case, I'm using a known character to separate data entries in my own file format. I renamed this to DATA_SEPARATOR to hopefully make it more clear.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
sidheart: This is so that things are in the correct order. Look at how SlidingWindow stores data in a LinkedList on calls to next(). Also see PersistableSlidingWindowTest#testNext(). I added an extra case to this function for extra sanity as well.

ktkrg: ha! We offerFirst() and pollLast(), it makes sense now.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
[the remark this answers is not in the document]
ktkrg: Thanks, not sure why I said File.separator, I was meaning to say `System.lineSeparator()`. As long as you use the same separator for reading and writing I'm ok.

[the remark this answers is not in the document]
sidheart: Changed to use `System.lineSeparator()`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
sidheart: I thought about it, and I see what you were getting at. I preserved the load in the constructor, but I removed the create file as it's unnecessary.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
```diff
@@ -0,0 +1,99 @@
...
+                                  Path filePath) {
+    super(slidingWindowSize, timeUnit);
+    this.pathToFile = filePath;
+    try {
+      if (Files.exists(pathToFile)) {
+        load(pathToFile);
+      }
+    } catch (IOException e) {
+      LOG.error("Couldn't create file {} to perform young generation tuning", pathToFile, e);
+      throw new IllegalArgumentException("Couldn't create or read a file at " + filePath);
```
vigyasharma, line 54: Do we really need to wrap this into an `IllegalArgumentException`? We could leave this as the checked `IOException`. It will ensure that callers handle it gracefully.

vigyasharma (approved):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
rguo-aws: file could be accidentally deleted by operator or even some other process(i.e. log pusher). should we create a new file in case IOException occurs ?

sidheart: I've changed this not to throw. If we can't load data on construction we'll just log an error.

A caller who is expecting data to be loaded can check if the window is empty.
A deleted/moved file issue will be resolved on the next call to write() as it includes an atomic move operation.

rguo-aws (approved):

ktkrg (approved):

PR #454, opened 2020-10-06
Add disk usage metrics

ktkrg:
*Fixes #:*

*Description of changes:*
This change adds disk space metrics.
Our current disk utilization metric only tracks the utilization rate(reads/s + writes/s) in the last 5 seconds.

In order for the under utilization RCA to detect under utilization of disks we need both low service rate and low usage of disk space.

Since the concept of a disk applies to the operating system more than to Java(which interfaces through file systems which abstract out the physical disk device construct), we measure the space available on a mounted partition. This change adds the following metrics: `Partition_TotalSpace`, `Partition_FreeSpace`, and `Partition_UsableFreeSpace` metrics to the metricsDB.

The metrics parses the output of `/proc/mounts` and ignores virtual devices and system related mountpoints.

*Tests:*
Tested on docker.
```
sqlite> select * from Partition_TotalSpace;
/usr/share/elasticsearch/data|/dev/vda1|62725623808.0|62725623808.0|62725623808.0|62725623808.0

sqlite> select * from Partition_FreeSpace;
/usr/share/elasticsearch/data|/dev/vda1|28551467008.0|28551467008.0|28551467008.0|28551467008.0
```

UTs incoming.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metrics_generator/linux/LinuxMountedPartitionMetricsGenerator.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metrics_generator/linux/LinuxMountedPartitionMetricsGenerator.java
@@ -1,0 +1,48 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.metrics_generator.linux;
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.collectors.MountedPartitionMetrics;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.hwnet.MountedPartitions;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.metrics_generator.MountedPartitionMetricsGenerator;
+import com.google.common.collect.ImmutableSet;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class LinuxMountedPartitionMetricsGenerator implements MountedPartitionMetricsGenerator {
+  private static final Map<String, MountedPartitionMetrics> suppliers = new HashMap<>();
+
+  @Override
+  public void addSample() {
+    MountedPartitions.addSample();
+  }
+
+  @Override
+  public Set<String> getAllMountPoints() {
+    return ImmutableSet.copyOf(suppliers.keySet());
+  }
+
+  public void addSupplier(final String mountPoint,
+      final MountedPartitionMetrics supplier) {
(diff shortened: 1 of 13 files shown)


vigyasharma (approved):
LGTM.
One general question - when do we plan to use ES node stats metrics for disk space vs these OS level metrics?

ktkrg:
> LGTM.
> One general question - when do we plan to use ES node stats metrics for disk space vs these OS level metrics?

Did you mean the `shard_size_in_bytes` metric?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metrics_generator/linux/LinuxMountedPartitionMetricsGenerator.java
```diff
@@ -0,0 +1,48 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.metrics_generator.linux;
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.collectors.MountedPartitionMetrics;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.hwnet.MountedPartitions;
```
sidheart, line 4: This depends on MountedPartitions, but MountedPartitions also depends on LinuxMountedPartitionMetricsGenerator. Is this ok? Can't you move the sampling logic from MountedPartitions into the addSample() function here?

sidheart (commented):
The logic in here is good but does the ES process always have access to /proc/mounts? This collector may fail in certain ODFE distributions.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metrics_generator/linux/LinuxMountedPartitionMetricsGenerator.java
ktkrg: So the generator calls addSample() in MountedPartitions class, where it updates its own map and sets it back in the generator. The idea is to have all logic related to profiling/measuring these metrics in one class separate from the generator. The other hardware metrics generators follow the same pattern as well. Didn't want to diverge as well.

> The logic in here is good but does the ES process always have access to /proc/mounts? This collector may fail in certain ODFE distributions.

In the distros we support, /proc/mounts has 777 access. So reading them should be fine.

sidheart (approved):

ktkrg:
Will add UTs in the next PR.

PR #457, opened 2020-10-06
Fix a bug in old gen policy

rguo-aws:
*Fixes #:*

*Description of changes:*
found a bug in end to end dev testing. The bucketization function had a bug and gave wrong bucket index. Fixing the bug and updating the corresponding UT to capture it. 

*Tests:*
tested on dev stack 

### DIFF
--- a/pa_config/rca.conf
+++ b/pa_config/rca.conf
@@ -93,46 +93,5 @@
     "ClusterTemperatureRca"
   ],
   "muted-deciders": [],
-  "muted-actions": [],
-
-  // Default Values are added in the config specific file as well
-  // Modify them as well whenever changing the default values (DeciderConfig.java)
-  "decider-config-settings": {
-    // Decreasing order of priority for the type of workload we can expect on the cluster.
-    // Priority order in the list goes from most expected to the lease expected workload type.
-    "workload-type": {
-      "priority-order": ["ingest", "search"]
-    },
-    // Decreasing order of priority for the type of cache which is expected to be consumed more.
-    // Priority order in the list goes from most used to the lease used cache type.
-    "cache-type": {
-      "priority-order": ["fielddata-cache", "shard-request-cache", "query-cache", "bitset-filter-cache"]
-    }
-  },
-
-  // Action Configurations
-  "action-config-settings": {
-//    // Cache Max Size bounds are expressed as %age of JVM heap size
-//    "cache-settings": {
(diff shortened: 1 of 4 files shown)


pa_config/rca.conf
```diff
@@ -93,46 +93,5 @@
     "ClusterTemperatureRca"
   ],
   "muted-deciders": [],
-  "muted-actions": [],
```
vigyasharma, line 96: We need to clean up these configs, `muted_actions` and `muted_deciders` are also only needed by master node. Can you create an issue for this?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/old_gen/LevelTwoActionBuilder.java
```diff
@@ -150,7 +150,7 @@ private void actionPriorityForCache() {
    */
   private int bucketization(int lowerBound, int upperBound, int val, int bucketSize) {
     double step = (double) (upperBound - lowerBound) / (double) bucketSize;
-    return (int) ((double) val / step);
+    return (int) ((double) (val - lowerBound) / step);
```
vigyasharma, line 153: nit: can we rename value by `queueSize` or `size` or `resourceSize` ?

pa_config/rca_master.conf
```diff
@@ -109,20 +109,47 @@
   // Default Values are added in the config specific file as well
   // Modify them as well whenever changing the default values (DeciderConfig.java)
   "decider-config-settings": {
-    // Decreasing order of priority for the type of workload we can expect on the cluster.
-    // Priority order in the list goes from most expected to the lease expected workload type.
     "workload-type": {
-      "priority-order": ["ingest", "search"]
+      "prefer-ingest": true,
+      "prefer-search": true
```
sidheart, line 114: The other format seems much clearer to me. Why did we change it to this? What does this mean?

vigyasharma (approved):

pa_config/rca_master.conf
rguo-aws: Specifying the priority order as an array has two issues : 
1. we can't assign search and index operations with the same level of priority
2. we need maintain a function to parse all elements in array

sidheart (approved):

PR #461, opened 2020-10-09
Add bucketization logic (heap_usage) for cache and queue deciders

sruti1312:
*Fixes #:*

*Description of changes:*
Add bucketization logic (heap_usage) for cache and queue deciders

Created a bucket for heap usage percentage. 
Cache and queue deciders take an action only heap usage is in healthy_with_buffer state.

* Currently choosing 0.3, 0.6 and 0.9 as threshold for different usage buckets for heap_usage tunable.
These values are chosen as starting point. We can tune these values after some experimentations.
* Also, there is no difference in the action when state is UNDER_SIZED or HEALTHY_WITH_BUFFER. We can chose to alter the step size based on this. This will be the next step.

*Tests:*
unit tests

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/DecisionMakerConsts.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/DecisionMakerConsts.java
@@ -15,9 +15,20 @@
 
 package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker;
 
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.bucket.UsageBucket;
+import com.google.common.collect.ImmutableMap;
 import com.google.gson.JsonParser;
 
 public class DecisionMakerConsts {
+    public static final String HEAP_TUNABLE_NAME = "heap-usage";
+
+    public static final ImmutableMap<UsageBucket, Double> HEAP_USAGE_MAP =
+            ImmutableMap.<UsageBucket, Double>builder()
+                    .put(UsageBucket.UNDER_UTILIZED, 30.0)
+                    .put(UsageBucket.HEALTHY_WITH_BUFFER, 60.0)
+                    .put(UsageBucket.HEALTHY, 90.0)
+                    .build();
+
     public static final String CACHE_MAX_WEIGHT = "maximumWeight";
 
     public static final JsonParser JSON_PARSER = new JsonParser();
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/DecisionMakerConsts.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/DecisionMakerConsts.java
@@ -22,6 +22,8 @@ import com.google.gson.JsonParser;
 public class DecisionMakerConsts {
     public static final String HEAP_TUNABLE_NAME = "heap-usage";
 
+    // Currently choosing 0.3, 0.6 and 0.9 as threshold for different usage buckets for heap_usage tunable.
+    // These values are chosen as starting point. We can tune these values after some experimentation.
     public static final ImmutableMap<UsageBucket, Double> HEAP_USAGE_MAP =
             ImmutableMap.<UsageBucket, Double>builder()
                     .put(UsageBucket.UNDER_UTILIZED, 30.0)
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
@@ -15,18 +15,27 @@
 
 package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.deciders;
 
+import static com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.DecisionMakerConsts.HEAP_TUNABLE_NAME;
+import static com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.DecisionMakerConsts.HEAP_USAGE_MAP;
+
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.Action;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.ModifyCacheMaxSizeAction;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.util.HeapUtil;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.ResourceEnum;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.flow_units.ResourceFlowUnit;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.HotClusterSummary;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.HotNodeSummary;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.HotResourceSummary;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.bucket.BasicBucketCalculator;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.bucket.BucketCalculator;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.bucket.UsageBucket;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.util.InstanceDetails;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.collector.NodeConfigCache;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.BaseClusterRca;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.FieldDataCacheClusterRca;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.NodeKey;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.ShardRequestCacheClusterRca;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.util.NodeConfigCacheReaderUtil;
 import com.google.common.collect.ImmutableMap;
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -46,6 +55,7 @@ public class CacheHealthDecider extends Decider {
   private final ImmutableMap<ResourceEnum, BaseClusterRca> cacheTypeBaseClusterRcaMap;
 
   List<ResourceEnum> modifyCacheActionPriorityList = new ArrayList<>();
+  private BucketCalculator heapUsageBucketCalculator;
   private int counter = 0;
 
   public CacheHealthDecider(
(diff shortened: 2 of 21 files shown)


sruti1312:
``` 
Test nodeRoleChange FAILED (2s)
FAILURE: Executed 455 tests in 39m 12s (1 failed)
```

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/DecisionMakerConsts.java
```diff
@@ -15,9 +15,22 @@
...
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.bucket.UsageBucket;
+import com.google.common.collect.ImmutableMap;
 import com.google.gson.JsonParser;
 
 public class DecisionMakerConsts {
+    public static final String HEAP_TUNABLE_NAME = "heap-usage";
+
+    // Currently choosing 0.3, 0.6 and 0.9 as threshold for different usage buckets for heap_usage tunable.
+    // These values are chosen as starting point. We can tune these values after some experimentation.
+    public static final ImmutableMap<UsageBucket, Double> HEAP_USAGE_MAP =
```
yojs, line 27: maybe we can call it `DEFAULT_HEAP_USAGE_MAP` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -92,13 +108,22 @@ private void configureActionPriority() {
...
     Action action = null;
 
-    for (String actionName : actionsByUserPriority) {
-      action =
-        getAction(actionName, esNode, threadPool, true);
-      if (action != null) {
-        break;
+    // there is no difference in the action when state is UNDER_SIZED or HEALTHY_WITH_BUFFER.
+    // TODO: Tune action parameters based on the state.
+    if (heapUsageBucketCalculator.compute(heapUsedPercent).equals(UsageBucket.HEALTHY_WITH_BUFFER)
```
yojs, line 119: Instead of calling it twice, `heapUsageBucketCalculator.compute(heapUsedPercent)` should we store it in a local variable ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -92,13 +108,22 @@ private void configureActionPriority() {
...
+                || heapUsageBucketCalculator.compute(heapUsedPercent).equals(UsageBucket.UNDER_UTILIZED)) {
+      for (String actionName : actionsByUserPriority) {
+        action =
+                getAction(actionName, esNode, threadPool, true);
```
yojs, line 123: maybe a line break is not needed and lines `122` and `123` can be merged ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/util/HeapUtil.java
```diff
@@ -0,0 +1,22 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.util;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class HeapUtil {
+    private static final Logger LOG = LogManager.getLogger(HeapUtil.class);
+
+    public static double getHeapUsedPercentage(final Long heapMaxInBytes, final Long heapUsedInBytes) {
```
yojs, line 9: I am thinking should we make the arguments primitive type and avoid the null check ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/summaries/ResourceUtil.java
```diff
@@ -35,6 +35,9 @@
   public static final Resource HEAP_MAX_SIZE = Resource.newBuilder()
           .setResourceEnum(ResourceEnum.HEAP)
           .setMetricEnum(MetricEnum.HEAP_MAX).build();
+  public static final Resource HEAP_USAGE = Resource.newBuilder()
+          .setResourceEnum(ResourceEnum.HEAP)
+          .setMetricEnum(MetricEnum.HEAP_USAGE).build();
   public static final Resource OLD_GEN_HEAP_USAGE = Resource.newBuilder()
```
yojs, line 41: This is a good change but I am thinking since we already have the `OLD_GEN_HEAP_USAGE` can we just use that ? That would mean we can get away with touching less files. 
The thing about accounting for full heap is that, not all the bytes are active and some of them will be reclaimed in the next gc. thoughts ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
```diff
@@ -130,7 +146,18 @@ private void configureModifyCacheActionPriority() {
...
+            NodeConfigCacheReaderUtil.readHeapMaxSizeInBytes(nodeConfigCache, esNode),
+            NodeConfigCacheReaderUtil.readHeapUsageInBytes(nodeConfigCache, esNode));
+
+    // there is no difference in the action when state is UNDER_SIZED or HEALTHY_WITH_BUFFER.
+    // TODO: Tune action parameters based on the state.
+    if (heapUsageBucketCalculator.compute(heapUsedPercent).equals(UsageBucket.HEALTHY_WITH_BUFFER)
+                || heapUsageBucketCalculator.compute(heapUsedPercent).equals(UsageBucket.UNDER_UTILIZED)) {
+      return getAction(ModifyCacheMaxSizeAction.NAME, esNode, cacheType, true);
+    }
+    return null;
```
yojs, line 160: Do we also want to suggest action to decrease the size if we are unhealthy ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
```diff
@@ -130,7 +146,18 @@ private void configureModifyCacheActionPriority() {
...
   private Action computeBestAction(final NodeKey esNode, final ResourceEnum cacheType) {
-    return getAction(ModifyCacheMaxSizeAction.NAME, esNode, cacheType, true);
+    final NodeConfigCache nodeConfigCache = getAppContext().getNodeConfigCache();
+    double heapUsedPercent = HeapUtil.getHeapUsedPercentage(
+            NodeConfigCacheReaderUtil.readHeapMaxSizeInBytes(nodeConfigCache, esNode),
+            NodeConfigCacheReaderUtil.readHeapUsageInBytes(nodeConfigCache, esNode));
+
+    // there is no difference in the action when state is UNDER_SIZED or HEALTHY_WITH_BUFFER.
+    // TODO: Tune action parameters based on the state.
+    if (heapUsageBucketCalculator.compute(heapUsedPercent).equals(UsageBucket.HEALTHY_WITH_BUFFER)
```
vigyasharma, line 156: nit: use local variable for `heapUsageBucketCalculator.compute(heapUsedPercent)`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
vigyasharma, line 160: Let us add an info log and add operational metrics for when the action should've been emitted but was punted due to heap utilization.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -92,13 +108,22 @@ private void configureActionPriority() {
...
 
-    for (String actionName : actionsByUserPriority) {
-      action =
-        getAction(actionName, esNode, threadPool, true);
-      if (action != null) {
-        break;
+    // there is no difference in the action when state is UNDER_SIZED or HEALTHY_WITH_BUFFER.
+    // TODO: Tune action parameters based on the state.
+    if (heapUsageBucketCalculator.compute(heapUsedPercent).equals(UsageBucket.HEALTHY_WITH_BUFFER)
+                || heapUsageBucketCalculator.compute(heapUsedPercent).equals(UsageBucket.UNDER_UTILIZED)) {
```
vigyasharma, line 120: same as cache, logs and metrics for when actions get punted due to heap util

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/util/HeapUtil.java
```diff
@@ -0,0 +1,22 @@
...
+        double heapUsedPercent = 0;
+
+        if (heapMaxInBytes != null && heapMaxInBytes > 0 && heapUsedInBytes != null && heapUsedInBytes > 0) {
+            heapUsedPercent = ((double) heapUsedInBytes / heapMaxInBytes) * 100;
+        }
+
+        if (!Double.isNaN(heapUsedPercent)) {
```
vigyasharma, line 16: when can this be NaN? wont the check above prevent that.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/util/HeapUtil.java
vigyasharma: yes, just need to avoid the 0/0 case.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCollector.java
```diff
@@ -101,6 +105,13 @@ private void collectHeapStats(MetricFlowUnit heapMax) {
     collectAndPublishMetric(ResourceUtil.YOUNG_GEN_MAX_SIZE, edenMaxSize + (2 * survivorMaxSize));
   }
 
+  private void collectHeapUsageStats(MetricFlowUnit heapUsed) {
```
vigyasharma, line 108: How do we decide b/w putting something in the NodeConfigCollector (or NodeConfigCache) v/s creating a node level and cluster level rca for it? Both are ways to get values to the master node deciders.

Up until now, we were using NodeConfigCache only for configuration values passed as a metric reading - e.g. queue capacity, heap max size, cache max size. Heap usage is a digression from that pattern.
I'm not saying this is wrong, but we should have some structure/design so as to avoid making the NodeConfigCollector as one big blob of all metrics.

IMO, it makes sense to use NodeConfigCollector only for fixed config values (like we have up until now), and move things like usage metrics (heap usage) to node and cluster level RCAs. This leaves the door open for any future post-processing we might want, e.g. returning the avg/min/ewma of heap usage over a sliding window rather than exact last value.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
```diff
@@ -59,10 +69,16 @@ public CacheHealthDecider(
...
-            .put(ResourceEnum.SHARD_REQUEST_CACHE, shardRequestCacheClusterRca)
-            .put(ResourceEnum.FIELD_DATA_CACHE, fieldDataCacheClusterRca)
-            .build();
+            ImmutableMap.<ResourceEnum, BaseClusterRca>builder()
+                    .put(ResourceEnum.SHARD_REQUEST_CACHE, shardRequestCacheClusterRca)
+                    .put(ResourceEnum.FIELD_DATA_CACHE, fieldDataCacheClusterRca)
+                    .build();
+    this.heapUsageBucketCalculator = new BasicBucketCalculator(HEAP_USAGE_MAP);
+    try {
+      this.heapUsageBucketCalculator = rcaConf.getBucketizationSettings(HEAP_TUNABLE_NAME);
```
vigyasharma, line 78: I am thinking that this bucket calculator should be a part of the rca and not the decider (also see comment about moving HeapUsage to an rca instead of NodeCofigCollector below).

This is because these bucketization values are closely linked to the rca or metric, rather than the decider. Suppose there are 3 different metrics that are being bucketized, say young-gen-util, old-gen-util and cpu-util. If decider applies bucketization, then each decider must remember the `tunableName` for each bucketization rcaconf setting. Whereas, if rcas apply it, they need to just remember their own `tunableName` values.

We should just pass rcaConf to the rcas and let them return the usage buckets. They can also chose the bucket calculator relevant to them, and implement some interface (`Bucketizable` ?) which exposes a `getUsageBucket` method.

PR #463, opened 2020-10-09
Make JvmActionsAlarmMonitor persistable

sidheart:
*Fixes #:* N/A

*Description of changes:*
- BucketizedSlidingWindow now extends PersistableSlidingWindow which
means it can write and load window data from a file on disk
- JvmActionsAlarmMonitor has a new constructor that will allow its data
to be written to and loaded from the disk

*Tests:* 
- PersistableSlidingWindowTest
- JvmActionsAlarmMonitorTest

*If new tests are added, how long do the new ones take to complete*: Negligible

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
@@ -23,11 +23,11 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Instant;
 import java.util.Iterator;
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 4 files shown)


vigyasharma (dismissed):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/framework/api/aggregators/PersistableSlidingWindow.java
```diff
@@ -94,5 +110,6 @@ protected synchronized void write() throws IOException {
     writer.flush();
```
ktkrg, line 110: You need to close the stream here after flushing. You can use a try-with-resources block so that it gets autoclosed after using.

sidheart: Good catch, done

ktkrg (approved):

vigyasharma (approved):

ktkrg:
SpotBugs issue(UrF) in SlidingWindow will be addressed in the next PR.

PR #465, opened 2020-10-13
Using to buckets to suppress queue length increase action

yojs:
*Fixes #:* #467

*Description of changes:* The Queue And Cache Deciders generates the actions to increase the Queue or Cache size if there are no errors. But this can have consequences if the Jvm Heap is already contended. This patch adds a check to make sure that the old gen heap occupancy after GC is underutilized or healthy and more of it can be consumed. The buckets to determine the usage limits are obtained from the `rca.conf`. If rca.conf does not contain them, then the defaults in the `HeapBasedDecider::DEFAULT_HEAP_USAGE_THRESHOLDS`.

*Tests:*
- Added unit tests to check the `canUseMoreHeap()` works as expected.
- Added an integration test. 

` Test testNoCapacityIncreaseOnUnHealthy PASSED (1m 59s)`

### DIFF
--- a/pa_config/rca.conf
+++ b/pa_config/rca.conf
@@ -93,5 +93,12 @@
     "ClusterTemperatureRca"
   ],
   "muted-deciders": [],
-  "muted-actions": []
+  "muted-actions": [],
+  "bucketization": {
+    "old-gen": {
+      "UNDER_UTILIZED": 10.0,
+      "HEALTHY_WITH_BUFFER": 30.0,
+      "HEALTHY": 70.0
+    }
+  }
 }
--- a/pa_config/rca.conf
+++ b/pa_config/rca.conf
@@ -97,8 +97,8 @@
   "bucketization": {
     "old-gen": {
       "UNDER_UTILIZED": 10.0,
-      "HEALTHY_WITH_BUFFER": 30.0,
-      "HEALTHY": 70.0
+      "HEALTHY_WITH_BUFFER": 60.0,
+      "HEALTHY": 80.0
     }
   }
 }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
@@ -15,6 +15,7 @@
 
 package com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.deciders;
 
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.PerformanceAnalyzerApp;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.Action;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.decisionmaker.actions.ModifyCacheMaxSizeAction;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.ResourceEnum;
@@ -22,6 +23,7 @@ import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.HotClusterSummary;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.HotNodeSummary;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.HotResourceSummary;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.metrics.RcaRuntimeMetrics;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.util.InstanceDetails;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.HighHeapUsageClusterRca;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.BaseClusterRca;
@@ -135,6 +137,9 @@ public class CacheHealthDecider extends HeapBasedDecider {
     Action action = null;
     if (canUseMoreHeap(esNode)) {
       action = getAction(ModifyCacheMaxSizeAction.NAME, esNode, cacheType, true);
+    } else {
+      PerformanceAnalyzerApp.RCA_RUNTIME_METRICS_AGGREGATOR.updateStat(
+          RcaRuntimeMetrics.NO_INCREASE_ACTION_SUGGESTED, NAME + ":" + esNode.getHostAddress(), 1);
     }
     return action;
   }
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
@@ -23,6 +23,7 @@ import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.HotNodeSummary;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.summaries.HotResourceSummary;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.util.InstanceDetails;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.HighHeapUsageClusterRca;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.BaseClusterRca;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.FieldDataCacheClusterRca;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.store.rca.cluster.NodeKey;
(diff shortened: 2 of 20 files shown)


yojs:
` Test nodeRoleChange FAILED (2s)`

`FAILURE: Executed 455 tests in 42m 36s (1 failed)`

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
```diff
@@ -130,7 +134,14 @@ private void configureModifyCacheActionPriority() {
    * signals going forward.
    */
   private Action computeBestAction(final NodeKey esNode, final ResourceEnum cacheType) {
-    return getAction(ModifyCacheMaxSizeAction.NAME, esNode, cacheType, true);
+    Action action = null;
+    if (canUseMoreHeap(esNode)) {
+      action = getAction(ModifyCacheMaxSizeAction.NAME, esNode, cacheType, true);
+    } else {
+      PerformanceAnalyzerApp.RCA_RUNTIME_METRICS_AGGREGATOR.updateStat(
```
khushbr, line 141: I am wondering if we should also add a log statement here in addition to the metric. Thoughts?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/HeapBasedDecider.java
```diff
@@ -0,0 +1,77 @@
...
+   * sustain more live objects in it without de-gradation. What is an acceptable heap usage limit to determine this, comes from the
+   * bucketization object in rca.conf. We compare the oldGen usage percent reported by the HighHeapUsage RCA to determine that.
+   *
+   * @param esNode The EsNode we are trying to make a decision for.
+   * @return return if the OldGen heap is under-utilized or healthy and yet more can be consumed, return true; or false otherwise.
+   */
+  protected boolean canUseMoreHeap(NodeKey esNode) {
+    // we add action only if heap is under-utilized or healthy and yet more can be consumed.
+    for (ResourceFlowUnit<HotClusterSummary> clusterSummary : highHeapUsageClusterRca.getFlowUnits()) {
+      if (clusterSummary.hasResourceSummary()) {
```
khushbr, line 46: should the check instead be ` clusterSummary.getSummary()` ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/HeapBasedDecider.java
```diff
@@ -0,0 +1,77 @@
...
+                } else {
+                  return false;
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+    return true;
```
khushbr, line 75: There are multiple level of nesting here with `for-if-for-if`, can we break down this function into two or more or use lambda ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -93,13 +99,16 @@ private void configureActionPriority() {
    */
   private Action computeBestAction(NodeKey esNode, ResourceEnum threadPool) {
     Action action = null;
-
-    for (String actionName : actionsByUserPriority) {
-      action =
-        getAction(actionName, esNode, threadPool, true);
-      if (action != null) {
-        break;
+    if (canUseMoreHeap(esNode)) {
```
khushbr, line 102: Indentation looks wrong here.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/HeapBasedDecider.java
yojs: `hasSummary` is that check. It checks if the summary is `null`.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/CacheHealthDecider.java
yojs: We don't want to emit the log continuously and in a cluster with jvm pressure, we will emit it quite a few times.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
yojs:
```java
  private Action computeBestAction(NodeKey esNode, ResourceEnum threadPool) {
    Action action = null;
    if (canUseMoreHeap(esNode)) {
      for (String actionName : actionsByUserPriority) {
        action = getAction(actionName, esNode, threadPool, true);
        if (action != null) {
          break;
        }
      }
    } else {
      PerformanceAnalyzerApp.RCA_RUNTIME_METRICS_AGGREGATOR.updateStat(
          RcaRuntimeMetrics.NO_INCREASE_ACTION_SUGGESTED, NAME + ":" + esNode.getHostAddress(), 1);
    }
    return action;
  }
```

yojs: I think the deletion and addition makes it look like it in the diff. This is code as it copied here. Can you check if the indentation looks right to you ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
```diff
@@ -93,13 +99,16 @@ private void configureActionPriority() {
...
+      for (String actionName : actionsByUserPriority) {
+        action = getAction(actionName, esNode, threadPool, true);
+        if (action != null) {
+          break;
+        }
       }
+    } else {
+      PerformanceAnalyzerApp.RCA_RUNTIME_METRICS_AGGREGATOR.updateStat(
+          RcaRuntimeMetrics.NO_INCREASE_ACTION_SUGGESTED, NAME + ":" + esNode.getHostAddress(), 1);
```
aditjind, line 111: Should we differentiate between an increase in the QueueIncreaseAction and Cache Increase Action?

pa_config/rca.conf
```diff
@@ -93,5 +93,12 @@
     "ClusterTemperatureRca"
   ],
   "muted-deciders": [],
-  "muted-actions": []
+  "muted-actions": [],
+  "bucketization": {
+    "old-gen": {
+      "UNDER_UTILIZED": 10.0,
```
aditjind, line 99: Add a comment saying Default Configuration Values are maintained at 2 places. In the code as well as here. We might miss updating at one place if we want to change the defaults in future.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/HeapBasedDecider.java
```diff
@@ -0,0 +1,77 @@
...
+    for (ResourceFlowUnit<HotClusterSummary> clusterSummary : highHeapUsageClusterRca.getFlowUnits()) {
+      if (clusterSummary.hasResourceSummary()) {
+        for (HotNodeSummary nodeSummary : clusterSummary.getSummary().getHotNodeSummaryList()) {
+          NodeKey thisNode = new NodeKey(nodeSummary.getNodeID(), nodeSummary.getHostAddress());
+          if (thisNode.equals(esNode)) {
+            for (HotResourceSummary hotResourceSummary : nodeSummary.getHotResourceSummaryList()) {
+              Resource resource = hotResourceSummary.getResource();
+              if (resource.getResourceEnum() == DECIDING_HEAP_RESOURCE_TYPE) {
+                double oldGenUsedRatio = hotResourceSummary.getValue();
+                double oldGenUsedPercent = oldGenUsedRatio * 100;
```
aditjind: Do we really need this extra variable `oldGenUsedRatio`?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/HeapBasedDecider.java
yojs: I have broken the function into two. I will keep thinking about making the loop nicer to look at if I find one. We may not block on that though.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/HeapBasedDecider.java
yojs: I added that for readability. Stack variables are cheap.

pa_config/rca.conf
yojs: The configs are there in the rca.conf although they have the same value as the one we have by default is it gives people an idea of what is configurable just by looking at the rca.conf and not having to read the code. rca.conf takes precedence.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/QueueHealthDecider.java
yojs: Right, the `NAME ` part does that.

pa_config/rca.conf
aditjind: What I meant was lets say we want to change the values tomorrow and we just change it in the Conf file and when we fall back to defaults from the code we might fall back to the previous defaults which have been changed. The comment here helps while updating. I dont have a strong opinion regarding this. We can choose to not add comment as well.

aditjind (approved):

khushbr (approved):

PR #439, opened 2020-09-30
Add 128GB heap JVM decider

ktkrg:
*Fixes #:*

*Description of changes:*
This change adds a new `JvmScaleUpPolicy` and the associated RCAs.

What's in the PR:
* Suggesting 128GB heap if there is old gen contention.

Known issues being addressed:
* There is some code duplication around reading values like **heap usage after gc**, **old gen utilization percentage** between the `HighHeapUsageOldGenRca` and the two new RCAs introduced in this PR. I'm refactoring them to derive from a base `OldGenRca(heapused, heapMax, gcEvents)` so that the code calculate these are in one place instead of many.
* Still using the old sliding window for now. Will use one of Leaky Bucket algo or Persistable sliding window.

*Tests:*

*If new tests are added, how long do the new ones take to complete*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #475, opened 2020-10-16
Add missing upstream rca node for heap health decider

ktkrg:
*Fixes #:*
#476 
*Description of changes:*
Adds LargeHeapClusterRca - the cluster level RCA for detecting heap size increase opportunities - as an upstream to the HeapHealthDecider.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElasticSearchAnalysisGraph.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElasticSearchAnalysisGraph.java
@@ -61,7 +61,6 @@ import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.metrics.ThreadPool_QueueCapacity;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.metrics.ThreadPool_RejectedReqs;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.api.metrics.VersionMap_Memory;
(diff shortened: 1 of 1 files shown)


yojs (approved):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElasticSearchAnalysisGraph.java
```diff
@@ -200,7 +199,7 @@ public void construct() {
     HeapHealthDecider heapHealthDecider = new HeapHealthDecider(RCA_PERIOD, highHeapUsageClusterRca,
         largeHeapClusterRca);
     heapHealthDecider.addTag(TAG_LOCUS, LOCUS_MASTER_NODE);
-    heapHealthDecider.addAllUpstreams(Collections.singletonList(highHeapUsageClusterRca));
+    heapHealthDecider.addAllUpstreams(Arrays.asList(highHeapUsageClusterRca, largeHeapClusterRca));
```
vigyasharma, line 202: Did we add decider integ tests? How did they pass without this?

sidheart (approved):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/ElasticSearchAnalysisGraph.java
ktkrg: No, we don't have decider integ tests yet.

> How did they pass without this?

Not having it in the upstream list won't necessarily fail the integration test even if we had one. The scheduler can at worst get the scheduling wrong and have the decider run before the RCA(because they'll be executed in parallel based on the distance from the leaf nodes). This causes race conditions which may or may not fail in the integ test. This upstream dependency was present in the 128G PR: #439 though, so this wouldn't have been caught then. Maybe at/after merge?

PR #483, opened 2020-10-21
Adjust resizable queue lower bounds to align with the default capacity(200/1000) suggested by ES

rguo-aws:
*Fixes #:*

*Description of changes:*
Our queue related performance experiments indicate that downsizing of the write queue could potentially increase heap pressure when a lot of rejection occurs within a short amount of time. So this PR will adjust the default queue lower bound back to the default queue capacity used by ES. This would at least guarantee that our resizable queue does not introduce extra heap pressure when compared with the vanilla ES. 
So now for the write queue, the lower/upper bounds are [200/1000]
and for search queue, the lower/upper bounds are [1000/2000]

*Tests:*
update UT as well so that the UTs are independent of the actual values being set for queue's lower/upper bounds

n/a

### DIFF
--- a/pa_config/rca_idle_master.conf
+++ b/pa_config/rca_idle_master.conf
@@ -109,22 +109,49 @@
   // Default Values are added in the config specific file as well
   // Modify them as well whenever changing the default values (DeciderConfig.java)
   "decider-config-settings": {
-    // Decreasing order of priority for the type of workload we can expect on the cluster.
-    // Priority order in the list goes from most expected to the lease expected workload type.
     "workload-type": {
-      "priority-order": ["ingest", "search"]
+      "prefer-ingest": true,
+      "prefer-search": true
     },
     // Decreasing order of priority for the type of cache which is expected to be consumed more.
     // Priority order in the list goes from most used to the lease used cache type.
     "cache-type": {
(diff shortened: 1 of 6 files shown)


sidheart (approved):
lgtm

pa_config/rca_idle_master.conf
```diff
@@ -109,22 +109,49 @@
...
-      "field-data-cache-upper-bound" : 0.4,
-      "shard-request-cache-upper-bound" : 0.05
+    "old-gen-decision-policy-config": {
+      "queue-bucket-size" : 20,
+      "old-gen-threshold-level-one" : 0.6,
+      "old-gen-threshold-level-two" : 0.75,
+      "old-gen-threshold-level-three" : 0.9
     }
   },
+  // Action Configurations
```
vigyasharma, line 128: Do we need to add all of these configurations? Wont just changing the default queue lower bounds suffice?

vigyasharma (approved):

pa_config/rca_idle_master.conf
rguo-aws: We don't have to add all configurations in the rca conf but I remember Joydeep once asked me to explicitly add all rca conf into the file so that people who are not familiar with the settings can easily understand and set it

PR #491, opened 2020-10-23
Fix total memory being read incorrectly

ktkrg:
*Fixes #:*

*Description of changes:*
Read the contents of /proc/meminfo to get total memory on the instance instead of JVM Runtime object which reads total memory incorrectly.

*Tests:*
Tested on an ec2 instance.


sidheart (approved):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/util/MemInfoParser.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    return -1;
+  }
+
+  private static long extractTotalMemory(final String memLine) {
+    String[] parsedLine = memLine.trim().replaceAll("\\s+", " ").split(" ");
+    if (parsedLine.length != 3) {
+      return -1;
+    }
```
khushbr, line 54: Can we add a comment around what this regex is trying to do  and why we are returning if length of parsedLine is not 3

khushbr (approved):

Issue #505, opened 2020-10-31
Actions are not getting muted as expected when disabled through overrides API


ktkrg:
When an action is disabled through the overrides API, we see that the action is still kept enabled and evaluates normally when it should have been disabled and be marked as muted.

PR #506, opened 2020-10-31
Fix actions not muting

ktkrg:
*Fixes #:*
#505 

*Description of changes:*
This PR fixes the bug where actions that were overridden to be disabled do not end up being marked as muted. The reason for this behavior is because the RCA controller, while updating the list of muted components, updates the graph nodes directly for RCAs, and deciders, it updates the list of muted actions in the AppContext.

However, the AppContext updated by the RcaController and the AppContext used by the RcaScheduler are different. The Scheduler only gets a snapshot of the AppContext during its initialization as it uses that AppContext to determine staleness, and never gets an updated AppContext till the scheduler is restarted.

This causes the update to go missing in the AppContext used by the RcaScheduler which eventually manifests as an unmuted action(when muting an action is involved). The fix basically exposes a new method from the scheduler to get the updated list of muted actions that it should pass down to the Actions through its AppContext.

*Tests:*
Unit tests, tested on docker with HeapSizeIncreaseAction.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
@@ -30,7 +30,6 @@ import com.amazon.opendistro.elasticsearch.performanceanalyzer.net.NetClient;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.net.NetServer;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.exceptions.MalformedConfig;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.core.ConnectedComponent;
-import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.core.MetricsDBProvider;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.core.Queryable;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.core.RcaConf;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.core.Stats;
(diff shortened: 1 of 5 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
```diff
@@ -410,7 +409,15 @@ private boolean updateMutedComponents() {
...
       Stats.getInstance().updateMutedGraphNodes(graphNodesForMute);
+      // We need to update muted actions in two places. One is in the controller which could
+      // potentially serve as a snapshot when creating a new rca scheduler, and the other place
+      // is in the rca scheduler itself. The scheduler maintains a snapshot of the appcontext(see
+      // the comment above near scheduler instantiation) that it uses to determine staleness. It
+      // also happens to be the appcontext that will be passed to the graph nodes.
       appContext.updateMutedActions(actionsForMute);
+      if (rcaScheduler != null) {
+        rcaScheduler.updateAppContextWithMutedActions(actionsForMute);
+      }
```
yojs, line 420: This fixes this issue but we can run into such inconsistencies where changes may not reflect until a scheduler restart. Can the scheduler use a reference of the AppContext of RcaController instead of a copy ?

ktkrg: That was my first thought as well. But I saw this comment on L215 and decided to go this way.
> // RcaScheduler should be started with a snapshot of the AppContext as RcaController
      // monitors it for stale state and always restarts the scheduler if it finds its state
      // stale.

I think you have the best context regarding AppContext usage, if this comment is stale, then I can go ahead and use the same AppContext reference in both the controller and the scheduler.

yojs: We will have to handle the case of changed role for the RCAScheduler. we can do the cleanup for a follow up PR. This is a bug fix, so I won't block this on the proper fix.

yojs (approved):

sidheart (approved):

PR #514, opened 2020-11-16
Add v1 cortisol code

ktkrg:
*Fixes #:*

*Description of changes:*
Simple load generator that ingests documents at a given rate.

*Tests:*

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tools/cortisol/IngestTask.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tools/cortisol/IngestTask.java
@@ -1,0 +1,172 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.tools.cortisol;
+
+import org.apache.http.HttpHost;
+import org.elasticsearch.action.ActionListener;
+import org.elasticsearch.action.bulk.BulkItemResponse;
+import org.elasticsearch.action.bulk.BulkRequest;
+import org.elasticsearch.action.bulk.BulkResponse;
+import org.elasticsearch.action.index.IndexRequest;
+import org.elasticsearch.client.*;
+import org.elasticsearch.client.indices.CreateIndexRequest;
+import org.elasticsearch.client.indices.CreateIndexResponse;
+import org.elasticsearch.common.xcontent.XContentType;
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class IngestTask implements Runnable {
+    private static final String CORTISOL_PREFIX = "cort-";
+    private final String endpoint;
+    private final int port;
+    private final BulkLoadParams params;
+    private final long endTime;
+    private final RestHighLevelClient client;
+    private final Random random;
+    private final String indexName;
+    private final TokenBucket tokenBucket;
+    private final AtomicInteger requestCount;
+    private final AtomicInteger successCount;
+    private final AtomicInteger failureCount;
+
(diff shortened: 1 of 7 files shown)


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tools/cortisol/IngestTask.java
```diff
@@ -0,0 +1,172 @@
...
+            this.maxTokens = maxTokens;
+            this.ses.schedule(this::refill, 1, TimeUnit.SECONDS);
+        }
+
+        public int take() {
+            lock.lock();
+            try {
+                while (nTokens == 0) {
+                    hasTokens.await();
+                }
```
sidheart, line 151: Does this need to be in a while loop?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tools/cortisol/IngestTask.java
```diff
@@ -0,0 +1,172 @@
...
+
+            @Override
+            public void onFailure(Exception e) {
+                failureCount.addAndGet(bulkDocs.size());
+            }
+        });
+        requestCount.addAndGet(bulkDocs.size());
+    }
+
+    private class TokenBucket {
```
sidheart, line 133: You should add a comment to this class since it functions slightly different than a standard token bucket. I believe this lets you make at most nTokens concurrent take() calls every second.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tools/cortisol/IngestTask.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+
+        private void refill() {
+            lock.lock();
+            try {
+                nTokens = maxTokens;
+                hasTokens.signal();
+            } finally {
+                lock.unlock();
+            }
```
sidheart, line 169: Why doesn't the take() method just replenish the bucket by 1 once it's done?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tools/cortisol/IngestTask.java
```diff
@@ -0,0 +1,172 @@
...
+                failureCount.addAndGet(bulkDocs.size());
+            }
+        });
+        requestCount.addAndGet(bulkDocs.size());
+    }
+
+    private class TokenBucket {
+        private final Lock lock = new ReentrantLock();
+        private final Condition hasTokens = lock.newCondition();
+        private final ScheduledExecutorService ses = Executors.newSingleThreadScheduledExecutor();
```
sidheart, line 136: I think you can avoid having a scheduler here. Have calls to take() block, and make sure every take() refills what it took from the bucket once it's complete.

sidheart (changes requested):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tools/cortisol/IngestTask.java
ktkrg: The work done after taking a token is dependent on the time each bulk() takes. We won't be able to guarantee a constant request rate. Some more explanation is in the next comment.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tools/cortisol/IngestTask.java
ktkrg: take() already blocks with await() but I see what you mean here. The idea is to make sure that we make a fixed number of calls every second. If we made the call to makeBulkRequest() block and replenish the tokens once the requests are done, then we won't be able to guarantee a constant request rate. We don't want to replenish too soon(if bulk calls finish faster) or too late(if the bulk calls take longer to finish).

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tools/cortisol/IngestTask.java
ktkrg: Yes, to guard against spurious wake ups.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/tools/cortisol/IngestTask.java
ktkrg: Done.

PR #528, opened 2020-12-08
Update the metricsDB instance for the gauntlet tests without a scheduler restart

yojs:
*Fixes #:*

*Description of changes:*

Today the metricsDB instance serving the graph node can be updated but for the
update to take effect, it also restarts the Scheduler. The scheduler stores the
states of the RCA nodes in memory and therefore restarting it cases all transient
states to be lost. This change tries to make the chnage without causing a Scheduler
restart. In order to do this, we add a check in the SchedulerTask. Before every
run invokation, it checks if someone has asked for the metricsDB to be updated
if so, then it goes ahead and updates the reference for all Tasklets. This is only
for test purposes and should not happen in prod.

*Tests:*
It is tested by adding a placeholder test in the dev environment. The test is not checked in as it does not test a real feature and will add to the already painful build time. Very soon tests will be added that leverage this feature.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/RcaController.java
@@ -526,7 +526,8 @@ public class RcaController {
     return rcaConf;
   }
 
-  public void setDbProvider(Queryable dbProvider) {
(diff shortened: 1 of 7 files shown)


rguo-aws (approved):
thanks for making this change. LGTM !

ktkrg (approved):
LGTM

PR #534, opened 2020-12-11
Add more operational metrics for metrics processing

ricardolstephen:
*Fixes #:*
#535

*Description of changes:*
Add more operational metrics

*Tests:*
Launched a docker cluster and checked that the metrics were properly collected.

Negligible

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/BatchMetricsEnabledSampler.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/BatchMetricsEnabledSampler.java
@@ -1,3 +1,18 @@
+/*
+ *  Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License").
+ *  You may not use this file except in compliance with the License.
+ *  A copy of the License is located at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
 package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.samplers;
 
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.AppContext;
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
@@ -1,0 +1,96 @@
+/*
+ *  Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License").
+ *  You may not use this file except in compliance with the License.
+ *  A copy of the License is located at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed
+ *  on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.samplers;
+
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.AppContext;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.metricsdb.MetricsDB;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.framework.metrics.ReaderMetrics;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.stats.collectors.SampleAggregator;
+import com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.stats.emitters.ISampler;
+
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.nio.file.Paths;
+import java.util.Objects;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class MetricsDBFileSampler implements ISampler {
+  private static final Logger LOG = LogManager.getLogger(MetricsDBFileSampler.class);
+  private static final Path targetDirectoryPath;
+  private static final DirectoryStream.Filter<Path> metricsdbFilter;
+  private static final DirectoryStream.Filter<Path> metricsdbTarFilter;
+  private final AppContext appContext;
+
+  static {
+    String metricsdbPrefix = MetricsDB.getFilePrefix();
+    targetDirectoryPath = Paths.get(metricsdbPrefix).getParent();
+    // Matches the metricsdb prefix concatenated with at least 1 digit
+    // Example: /tmp/metricsdb_1607650180000
+    String metricsdbPattern = "regex:" + metricsdbPrefix + "\\d+";
+    // Matches the metricsdb prefix, less one character, concatenated with a suffix like ".tar.2020-12-10-17-35.gz"
+    // Example: /tmp/metricsdb.tar.2020-12-10-17-40.gz
+    String metricsdbTarPattern = "regex:" + metricsdbPrefix.substring(0, metricsdbPrefix.length() - 1)
+        + ".tar.\\d+-\\d+-\\d+-\\d+-\\d+.gz";
+    PathMatcher metricsdbMatcher = FileSystems.getDefault().getPathMatcher(metricsdbPattern);
+    PathMatcher metricsdbTarMatcher = FileSystems.getDefault().getPathMatcher(metricsdbTarPattern);
+    metricsdbFilter = metricsdbMatcher::matches;
+    metricsdbTarFilter = metricsdbTarMatcher::matches;
+  }
+
+  public MetricsDBFileSampler(final AppContext appContext) {
+    Objects.requireNonNull(appContext);
+    this.appContext = appContext;
+  }
+
+  @Override
+  public void sample(SampleAggregator sampleCollector) {
+    int numUncompressedMetricsdbFiles = 0;
+    long sizeUncompressedMetricsdbFiles = 0;
+    int numMetricsdbFiles = 0;
+    long sizeMetricsdbFiles = 0;
+
+    try (DirectoryStream<Path> metricsdbStream = Files.newDirectoryStream(targetDirectoryPath, metricsdbFilter);
+         DirectoryStream<Path> metricsdbTarStream = Files.newDirectoryStream(targetDirectoryPath, metricsdbTarFilter)) {
+      for (Path entry : metricsdbStream) {
+          sizeUncompressedMetricsdbFiles += Files.size(entry);
+          numUncompressedMetricsdbFiles += 1;
+      }
+      for (Path entry : metricsdbTarStream) {
+          sizeMetricsdbFiles += Files.size(entry);
+          numMetricsdbFiles += 1;
+      }
+    } catch (IOException e) {
+      // Exceptions can arise here if the streams contain a path whose underlying file is deleted before the path is
+      // consumed.
+      LOG.warn("Issue accessing metricsdb entries in {}", targetDirectoryPath);
+      return;
+    }
+
+    numMetricsdbFiles += numUncompressedMetricsdbFiles;
+    sizeMetricsdbFiles += sizeUncompressedMetricsdbFiles;
+
+    sampleCollector.updateStat(ReaderMetrics.METRICSDB_NUM_FILES, "", numMetricsdbFiles);
+    sampleCollector.updateStat(ReaderMetrics.METRICSDB_SIZE_FILES, "", sizeMetricsdbFiles);
+    sampleCollector.updateStat(ReaderMetrics.METRICSDB_NUM_UNCOMPRESSED_FILES, "", numUncompressedMetricsdbFiles);
+    sampleCollector.updateStat(ReaderMetrics.METRICSDB_SIZE_UNCOMPRESSED_FILES, "", sizeUncompressedMetricsdbFiles);
+  }
+}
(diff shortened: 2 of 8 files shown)


khushbr:
Can we add an Issue for this PR ?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
```diff
@@ -0,0 +1,83 @@
+package com.amazon.opendistro.elasticsearch.performanceanalyzer.rca.samplers;
```
khushbr, line 1: nit: Add License information.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
```diff
@@ -0,0 +1,83 @@
...
+        try {
+          sizeUncompressedMetricsdbFiles += Files.size(entry);
+          numUncompressedMetricsdbFiles += 1;
+        } catch (IOException e) {
+          LOG.error("Failed to access metricsdb file - {}", entry);
+        }
+      }
+    } catch (IOException e) {
+      LOG.error("Failed to access metricsdb directory - {}", targetDirectoryPath);
+    }
```
sruti1312, line 60: Can we make this logic into a util method that both metricsdbTarFilter and metricsdbFilter can use?

sruti1312 (changes requested):

sruti1312 (approved):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
ricardolstephen: I don't think that'll work well since we want it the logic to produce two variables (sizeUncompressedMetricsdbFiles and numUncompressedMetricsdbFiles). Also, the duplication here doesn't seem like too much to require factoring out the logic.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
```diff
@@ -406,13 +392,10 @@ private void emitMetrics(long currWindowStartTime) throws Exception {
 
     metricsDB.commit();
     metricsDBMap.put(prevWindowStartTime, metricsDB);
-    long metricsDBSize = new File(metricsDB.getDBFilePath()).length();
     PerformanceAnalyzerApp.READER_METRICS_AGGREGATOR.updateStat(
-        ReaderMetrics.METRICSDB_FILE_SIZE, "", metricsDBSize);
+        ReaderMetrics.METRICSDB_FILE_SIZE, "", new File(metricsDB.getDBFilePath()).length());
```
khushbr, line 396: Instead of creating a new file and getting its length, can we replace `new File(metricsDB.getDBFilePath()).length()` with `metricsDB.getDBFilePath().length()`?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @Override
+  public void sample(SampleAggregator sampleCollector) {
+    String metricsdbPrefix = MetricsDB.getFilePrefix();
+
+    Path targetDirectoryPath = Paths.get(metricsdbPrefix).getParent();
+
+    String metricsdbPattern = "regex:" + metricsdbPrefix + "\\d+";
+    String metricsdbTarPattern = "regex:" + metricsdbPrefix.substring(0, metricsdbPrefix.length() - 1)
+        + ".tar.\\d+-\\d+-\\d+-\\d+-\\d+.gz";
```
khushbr, line 38: Please add a comment here stating what this regex represents. 
Also, we should add UT [here](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/tree/700771e321fd5bde14d2921153e183f03b7c2901/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers) for checking this regex behavior.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
```diff
@@ -0,0 +1,83 @@
...
+    PathMatcher metricsdbMatcher = FileSystems.getDefault().getPathMatcher(metricsdbPattern);
+    PathMatcher metricsdbTarMatcher = FileSystems.getDefault().getPathMatcher(metricsdbTarPattern);
+    DirectoryStream.Filter<Path> metricsdbFilter = metricsdbMatcher::matches;
+    DirectoryStream.Filter<Path> metricsdbTarFilter = metricsdbTarMatcher::matches;
```
khushbr, line 42: Instead of performing a `FileSystems.getDefault().getPathMatcher(metricsdbPattern);`, can use a config to define the location of these files ?  We are performing something similar here : https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/blob/master/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java#L351

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
```diff
@@ -0,0 +1,83 @@
...
+        try {
+          sizeMetricsdbFiles += Files.size(entry);
+          numMetricsdbFiles += 1;
+        } catch (IOException e) {
+          LOG.error("Failed to access metricsdb file - {}", entry);
+        }
+      }
+    } catch (IOException e) {
+      LOG.error("Failed to access metricsdb directory - {}", targetDirectoryPath);
+    }
```
khushbr, line 73: Can we avoid the nested `try-catch` blocks here? Let's move both the `DirectoryStream`  object creation to one 'try-catch' and then have inner 'try-catch' as now. Also, update the variable name from `stream` to `metricsdbStream` and `metricsdbTarStream` resp.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
```diff
@@ -0,0 +1,83 @@
...
+      for (Path entry : stream) {
+        try {
+          sizeUncompressedMetricsdbFiles += Files.size(entry);
+          numUncompressedMetricsdbFiles += 1;
+        } catch (IOException e) {
+          LOG.error("Failed to access metricsdb file - {}", entry);
+        }
+      }
+    } catch (IOException e) {
+      LOG.error("Failed to access metricsdb directory - {}", targetDirectoryPath);
```
khushbr, line 59: Will this error log overwhelm the log file in any case? Can we somehow reduce the frequency, maybe club all the failures and put a single log message?

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
khushbr, line 42: Certain variables between Line 32-40 can be made class static variables, move them out of the `sample()` method.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
ricardolstephen: Sure

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/reader/ReaderMetricsProcessor.java
ricardolstephen: I don't think that would work. `getDBFilePath` actually returns a String, not a Path. Also, a Path doesn't have a method to retrieve the size of the underlying file.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
ricardolstephen: Not completely sure what you mean by "config" here. There currently a configurable prefix (an absolute path prefix) for these metricsdb files. We probably don't want to split this prefix into a directory and filename prefix if that's what you mean -- doesn't seem too compelling to take on this new addition just for these two uses at the moment.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
ricardolstephen: Not a huge fan of removing the nested try-catch -- I don't want issues with one or two files corrupting the whole sampling. But yes, will make the other changes.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
ricardolstephen: Sure, will try something like that.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
ricardolstephen: Sure

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
ricardolstephen: Verified the correct behavior of the regex manually. The ut's can be a little more ticky for this due to the use of private variables and static methods. Will raise another separate issue to add these uts #538.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
khushbr: nit: Please update the year to 2020, instead of 2019.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
khushbr: I refer to using a variable like `DB_FILE_PREFIX_PATH_CONF_NAME` used [here](https://github.com/opendistro-for-elasticsearch/performance-analyzer-rca/blob/master/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/metricsdb/MetricsDB.java#L351) and my concern was regarding the ` FileSystems.getDefault().getPathMatcher(metricsdbPattern);`. I see that you have addressed this in the latest code revision. 

```
String prefix = PluginSettings.instance().getSettingValue(DB_FILE_PREFIX_PATH_CONF_NAME, DB_FILE_PREFIX_PATH_DEFAULT);
```

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
khushbr: I still dont understand why we have nested try-catch here.  Nested try-catch should be used in case the `catch` or `finally` block throws an exception. We have only Log statements in catch block here. 

There should be Unit Tests and Integrations Tests written to check for all failure scenarios, nested try-catch is hardly the way.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/BatchMetricsEnabledSampler.java
```diff
@@ -1,3 +1,18 @@
+/*
```
khushbr, line 1: nit: Make the year 2020 instead of 2019

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
```diff
@@ -0,0 +1,118 @@
...
+  private static final Path targetDirectoryPath;
+  private static final DirectoryStream.Filter<Path> metricsdbFilter;
+  private static final DirectoryStream.Filter<Path> metricsdbTarFilter;
+  private final AppContext appContext;
+
+  static {
+    String metricsdbPrefix = MetricsDB.getFilePrefix();
+    targetDirectoryPath = Paths.get(metricsdbPrefix).getParent();
+    String metricsdbPattern = "regex:" + metricsdbPrefix + "\\d+";
+    String metricsdbTarPattern = "regex:" + metricsdbPrefix.substring(0, metricsdbPrefix.length() - 1)
```
khushbr, line 47: Can we add a comment here about this regex. I am not comfortable with having regex in the code without any UTs/ITs present for testing.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
ricardolstephen: Sure

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
ricardolstephen: Going to change this to use a blanket try-catch and simply return without sampling if any issues are found (say, if the stream contains a Path whose associated file is deleted in the middle of sampling).

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/BatchMetricsEnabledSampler.java
ricardolstephen: Sure

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/samplers/MetricsDBFileSampler.java
ricardolstephen: Sure, will add a UT to this

khushbr (approved):

sruti1312 (approved):

PR #539, opened 2020-12-15
Add basic retry mechanism for clients

ktkrg:
*Fixes #:*

*Description of changes:*
Adds a simple retry mechanism for the grpc clients to retry connection failures.

This only retries connection failures. A generic mechanism to handle application level failures and retry them will need more investigation into when a stream should be rebuilt vs rpc being retried without re-establishment and will be added in a subsequent PR.

*Tests:*
Tested on a docker container by changing port on one node and live debugging on another node to make sure 2 additional calls were made from the grpc client lib.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
@@ -16,7 +16,6 @@
 package com.amazon.opendistro.elasticsearch.performanceanalyzer.net;
 
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.CertificateUtils;
-import com.amazon.opendistro.elasticsearch.performanceanalyzer.core.Util;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.InterNodeRpcServiceGrpc;
 import com.amazon.opendistro.elasticsearch.performanceanalyzer.grpc.InterNodeRpcServiceGrpc.InterNodeRpcServiceStub;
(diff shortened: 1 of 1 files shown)


khushbr (approved):

yojs (approved):

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/net/GRPCConnectionManager.java
```diff
@@ -175,10 +173,12 @@ private int getPortFromHost(final InstanceDetails remoteHost) {
...
   private ManagedChannel buildInsecureChannel(final InstanceDetails remoteHost) {
-    return ManagedChannelBuilder.forAddress(
-            remoteHost.getInstanceIp().toString(),
-            getPortFromHost(remoteHost)
-    ).usePlaintext().build();
+    return ManagedChannelBuilder.forAddress(remoteHost.getInstanceIp().toString(),
+        getPortFromHost(remoteHost))
+                                .usePlaintext()
+                                .enableRetry()
+                                .maxRetryAttempts(MAX_RETRY_ATTEMPTS)
```
rguo-aws, line 180: do we also want to add deadline for rpc call ?

And I also notice that maxRetryAttempts() is marked as experimental. is it going to be an issue to use it inprod ?

ktkrg: Deadlines for RPC - We use streaming calls, and deadlines aren't applicable to long lived streaming calls. I don't think there is a simple way to enforce deadlines per message. If we were using a blocking call/unary rpc, then we can add a deadline for it.

Both the enableRetry() and maxRetryAttempts() are marked experimental because of a different reason - its dependency on service config apis. We will consume those when we implement application level retries.

PR #550, opened 2021-01-20
Add spotless plugin to gradle build and configure license header

jotok:
This PR adds the Spotless plugin to this repository and configures it to use a header file to set the license header for .java files. The file diff consists of the changes to build.gradle plus changes to source files to fix header formatting and to add headers that were missing.

The diff for the build.gradle file has some commented out lines. These will be uncommented in a follow-up PR. I'm doing this as a two-step process because I want the license headers to be updated for every source file, but I don't want to update the code formatting for every source file because of the size of the diff it would create.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,6 +32,7 @@ plugins {
     id "de.undercouch.download" version "4.0.4"
     id 'com.adarshr.test-logger' version '2.1.0'
     id 'org.gradle.test-retry' version '1.2.0'
+    id 'com.diffplug.spotless' version '5.8.2'
 }
 
 application {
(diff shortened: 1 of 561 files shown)


build.gradle
```diff
@@ -87,6 +88,24 @@ checkstyle {
...
+spotless {
+    java {
+        licenseHeaderFile(file('license-header'))
+
+        // only apply formatting rules to updated files
+        // ratchetFrom 'origin/master'
+
+        // googleJavaFormat()
+        // importOrder()
+        // removeUnusedImports()
```
aditjind, line 102: Should we also add `trimTrailingWhitespace` and `endWithNewline` in the follow up PR?

aditjind (dismissed):

nisheedh (approved):

build.gradle
jotok: Makes sense, I'll add it to the next PR.

jotok: I added these options (commented out) in the rebased commit.

jotok:
Rebase.

aditjind (approved):

sendkb (approved):

PR #562, opened 2021-02-27
Only emit modify cache actions if the cache actual usage is above lower bound

rguo-aws:
*Fixes #:*

*Description of changes:*
Modify cache actions nowadays will reduce cache max capacity regardless of the actual usage of each cache. This will make users confused if they receive notices on cache actions being published but does not use i.e. fielddata cache at all. 
So this PR will Take cache actual size into consideration and only emit modify cache actions if the cache actual usage is above lower bound

*Tests:*
testing in progress

    indicated in the file; or

    in the file; or

    it.


src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/rca/store/collector/NodeConfigCollector.java
```diff
@@ -48,17 +50,23 @@
...
   private int counter;
   private final HashMap<Resource, Double> configResult;
 
   public NodeConfigCollector(int rcaPeriod,
                              ThreadPool_QueueCapacity threadPool_queueCapacity,
                              Cache_Max_Size cacheMaxSize,
-                             Heap_Max heapMaxSize) {
+                             Heap_Max heapMaxSize,
+                             AggregateMetric fieldDataCacheSizeGroupByAggregation,
+                             AggregateMetric shardRequestCacheSizeGroupByAggregation) {
```
khushbr, line 64: Why did we choose to add the FieldData and Request cache usage to Node Config Cache ?
Both of these will keep changing frequently and will require frequent cache updated. Why not use the `Cache_FieldData_Size` and `Cache_Request_Size` metrics available to us ?

PR #585, opened 2021-03-26
Remove canWrite check

sruti1312:
*Fixes #:*

*Description of changes:*
Reader does not write any files to the "metrics-location" - Shared memory space between the reader and the writer. 

1. Removing this canWrite() check on the reader. 

*Tests:*

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
+++ b/src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
@@ -244,16 +244,14 @@ public class PluginSettings {
       LOG.info("Trying to create directory {}.", path);
     }
 
-    boolean valid = dictCreated && dict.isDirectory() && dict.canWrite();
+    boolean valid = dictCreated && dict.isDirectory();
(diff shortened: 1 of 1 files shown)


nisheedh (approved):
looks good to me.

src/main/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/config/PluginSettings.java
```diff
@@ -244,16 +244,14 @@ private static void validateOrCreateDir(String path) throws ConfigFatalException
       LOG.info("Trying to create directory {}.", path);
     }
 
-    boolean valid = dictCreated && dict.isDirectory() && dict.canWrite();
+    boolean valid = dictCreated && dict.isDirectory();
```
yojs, line 247: The `PluginSettings` class is statically initialized. So this check happens in the context of the writer and the reader. I agree that writable check is not required for the reader but this is a required check for the writer. So, if we want to remove this check, we have to selectively do this only for the reader.

yojs (changes requested):

PR #596, opened 2021-05-07
Modify JVM gen tuning policy and update unit test

hxiao608:
*Description of changes:*
Modify the JVM gen tuning policy -> computeIncrease function to prevent increasing young gen size if the ratio is not below 4:1. Update the Unit test.

*Tests:*
Unit test passed

    indicated in the file; or

    in the file; or

    it.

### DIFF
--- a/src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/JvmGenTuningPolicyTest.java
+++ b/src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/JvmGenTuningPolicyTest.java
@@ -156,6 +156,15 @@ public class JvmGenTuningPolicyTest {
     Assert.assertEquals(1, actions.size());
     Assert.assertTrue(actions.get(0) instanceof JvmGenAction);
     Assert.assertEquals(5, ((JvmGenAction) actions.get(0)).getTargetRatio());
+    mockCurrentRatio(5);
+    actions = policy.evaluate();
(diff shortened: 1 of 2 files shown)


khushbr (approved):

sruti1312 (approved):

src/test/java/com/amazon/opendistro/elasticsearch/performanceanalyzer/decisionmaker/deciders/jvm/JvmGenTuningPolicyTest.java
```diff
@@ -167,5 +176,19 @@ public void testEvaluate() {
     Assert.assertEquals(1, actions.size());
     Assert.assertTrue(actions.get(0) instanceof JvmGenAction);
     Assert.assertEquals(3, ((JvmGenAction) actions.get(0)).getTargetRatio());
+    mockCurrentRatio(5);
+    actions = policy.evaluate();
+    Assert.assertEquals(1, actions.size());
```
aditjind, line 181: We should avoid using multiple asserts separately in a single unit test. If the first one fails, the test would exit and the following asserts won't execute, leaving us with no information on whether the following asserts would also fail. Also I would suggest adding failure message string in the assertion statements as that would increase the readability of the code and would also help in future maintenance. This is true for all the assertions in the current test file.

Sample Statement for the added assertions: 

```
Assert.assertTrue("Reason as to what is the cause of this assertion failure", 
                1 == actions.size()
                && actions.get(0) instanceof JvmGenAction
                && 4 == ((JvmGenAction) actions.get(0)).getTargetRatio());
```

